PR #20, opened 2020-10-22
App engine config and Serve frontend from static build

jtuckerk:
Set up the Portal frontend to be deployable to Google App Engine.

This requires setting up the start scripts to build then serve the app from the static build directory. There may be better ways to do this, but I'm not familiar enough with these technologies to know what the right way to do this is.

The app is up and running at https://humanity-forward.uc.r.appspot.com/


ecaoile (approved):

Swimburger:
I think we can host the python API and front-end on the same App Engine.

PR #24, opened 2020-10-27
Appengine cd

jtuckerk:
adding continuous deployment github actions and app engine config scripts for prod and staging api and just the staging client.
I will create a separate PR to be merged to master for the CD for the prod client.


Swimburger (approved):
Looks good. 
You could define working-directory at a broader scope to not repeat the parameter in multiple steps.

ecaoile (approved):

trippersham (approved):

jtuckerk:
In a sad turn of events I have discovered that there's no way to safely use github actions for deployment. The TLDR is that once I store a secret in github, anyone with push access to any branch (every team member) can push changes to the actions that are used to deploy the apps, and therefore point the actions to any branch and deploy anything they want.

worse yet, anyone could directly extract the secret and have service account level access. I thought there would be ways to restrict github actions, but all of the restrictions apply only to external actions that are linked in. I haven't found a way to freeze the actions or limit who can edit them. :( 

I'll be sending out a new PR with just the appengine configs and new scripts that can be used by the people with app engine deploy permissions.

PR #27, opened 2020-10-28
Setup Dockerfile from google's node image

indiemini:
Simple dockerfile built from google's nodejs image.

Usage:
`docker build --tag volportal:0.5 .` (builds the image)
`docker run --publish 80:8080 --detach --name vp volportal:0.5` (runs the built image in the background with label "vp")
`docker stop vp` (stops the process gracefully)
`docker rm vp` (removes the image from the active process list)


ecaoile (approved):
Had a derp moment, but it works on my machine now.

PR #28, opened 2020-10-28
Add AppEngine configs and manual deploy scripts

jtuckerk:
My dreams of CD have been crushed. If I'm mistaken about my reasoning for the security flaws in Github actions please let me know and we can reopen the CD PR.

Copying my comment over from the close CD PR:

In a sad turn of events I have discovered that there's no way to safely use github actions for deployment. The gist is that once I store a secret in github, anyone with push access to any branch (every team member) can push changes to the actions that are used to deploy the apps, and therefore point the actions to any branch and deploy anything they want.

worse yet, anyone could directly extract the secret and have service account level access. I thought there would be ways to restrict github actions, but all of the restrictions apply only to external actions that are linked in. I haven't found a way to freeze the actions or limit who can edit them. :(


ecaoile (approved):
RIP the dream

trippersham (approved):
I'll be candid that I'm out of my depth on CD; approving so I'm not in the way

PR #32, opened 2020-10-31
WIP

jtuckerk:
Adding this PR as a place to comment/discuss


api/data_source_maps.py
```diff
@@ -0,0 +1,27 @@
+hf_events = {
+    'event_uuid': None,
```
jtuckerk, line 2: the id field in the airtable tables is a unique ID if you need one.

api/data_source_maps.py
```diff
@@ -0,0 +1,27 @@
...
+    'description': 'description',
+    'point_of_contact': None,
+    'sign_up_link': 'signup_link'
+}
+
+hf_volunteer_openings = {
+    'role_uuid': None,
+    'role_external_id': 'position_id',
+    'details_url': None,
+    'hero_image_url': 'team_photo',
```
jtuckerk, line 16: this are null for now. looks like the airtable field name changed. I've added myself an asana task to lock the fields that we're using https://app.asana.com/0/1196959085120745/1198988802197301/f

It looks like the image urls are private, so I'll need to do some work to put them up in a public cloud storage bucket for serving. That might be somewhat involved depending on what it takes to on the airtable end

jtuckerk: The images could also be added to GCS and the DB manually for now.

jtuckerk: nevermind: the api returns links that are publicly viewable, and they updated their api I think so we get a nice json object with different sizes of the image:

[
  {
    "id": "attCGRTyhpWMThp9K",
    "url": "https://dl.airtable.com/.attachments/95e5b108af4e82b7e09570040e28dece/1508d41e/pexels-photo-1089440.jpeg",
    "size": 97014,
    "type": "image/jpeg",
    "filename": "pexels-photo-1089440.jpeg",
    "thumbnails": {
      "full": {
        "url": "https://dl.airtable.com/.attachmentThumbnails/457e213f2f21843a8e2867deb418b594/049533c6",
        "width": 3000,
        "height": 3000
      },
      "large": {
        "url": "https://dl.airtable.com/.attachmentThumbnails/df5021592f491d3b7efce036918979c3/a7df6f71",
        "width": 769,
        "height": 512
      },
      "small": {
        "url": "https://dl.airtable.com/.attachmentThumbnails/5f9c4f3ba1268bfab72a1488a76a5a30/2dac1752",
        "width": 54,
        "height": 36
      }
    }
  }
]

api/data_source_maps.py
```diff
@@ -0,0 +1,27 @@
+hf_events = {
+    'event_uuid': None,
+    'event_external_id': 'event_id',
+    'details_url': None,
+    'start_datetime': 'start',
+    'end_datetime': 'end',
+    'description': 'description',
+    'point_of_contact': None,
+    'sign_up_link': 'signup_link'
+}
```
jtuckerk, line 10: we have event graphics for events: event_graphics

api/models.py
```diff
@@ -13,17 +15,40 @@ class Priority(Enum):
...
+    event_uuid: UUID = uuid4()
+    event_external_id: str
+    details_url: Optional[Url]
+    start_datetime: datetime
+    end_datetime: Optional[datetime]
+    description: Optional[MarkdownText]
+    point_of_contact: Optional[Person]
+    sign_up_link: Optional[Url]
+
+class Initiative(BaseModel):
```
jtuckerk, line 47: Should we create an initiatives table in airtable for this?

api/data_source_maps.py
jtuckerk: ah nevermind. I get it now. you're generating the uid in the service.

api/models.py
jtuckerk: image
I've started working on this. The result in the DB for the roles and events columns will be a list of IDs that map to the external_id fields as you've named them.
I'll work on communicating this setup to Ericka so that we can get her input or organizing and grouping things. If you think we won't have time for this, feel free to pull me back. 
I'm thinking of making the personal link stuff as simple as possible, by just dumping the requests into a DB and then manually emailing people their personalized links. I think that'll free up a good chunk of my time this weekend.

api/data_sources.py
```diff
@@ -45,49 +44,139 @@ def __init__(self, address: str, data_source_type: DataSourceType, **data) -> No
...
+datasets: Dict[Tuple[DataSource, str, Optional[Type]], 'Dataset'] = {}
+
 class Dataset(BaseModel):
     dataset_uuid: UUID = uuid4()
     data_source: DataSource
     dataset_key: str
-    mapped_model: Optional[Type]
-    model_key_map: Optional[Dict[str, Optional[str]]]
+    primary_key: str
+    linked_model: Optional[Type]
```
jtuckerk, line 54: Notes to self for tucker for tomorrow, but feel free to answer them if you're here and it's simple:
What is a linked model

api/data_sources.py
```diff
@@ -45,49 +44,139 @@ def __init__(self, address: str, data_source_type: DataSourceType, **data) -> No
...
+
+                        if is_iterable:
+                            args[model_param_name] = [instance for key in naive_args[model_param_name] if (instance := dataset_for_nested_model.get_linked_model_object_for_primary_key(key))] if naive_args[model_param_name] else None
+                        else:
+                            args[model_param_name] = dataset_for_nested_model.get_linked_model_object_for_primary_key(naive_args[model_param_name]) if naive_args[model_param_name] else None
+                    else:
+                        # if a dataset linked to this model does not exist, attempt to pass value directly to this model's constructor
+                        args[model_param_name] = naive_args[model_param_name]
+
+        return self.linked_model(**args)
```
jtuckerk, line 179: image

api/data_sources.py
trippersham: A linked model is the Model in our vol-portal data model that this data source is supposed to sync to. This supports a theoretical {location of the external dataset, model it's supposed to sync to, simple field mapping} interface that we would expose to organizers setting up their own data connections for their own chapter/org.

I should add some docstrings before this gets merged.

PR #34, opened 2020-11-03
Initiatives API

trippersham:
API in "good enough" state for v0. to support Initiatives, VolunteerRoles, and VolunteerEvents


trippersham:
@jtuckerk could you help me diagnose that linting error in the build pipeline?

jtuckerk:
> @jtuckerk could you help me diagnose that linting error in the build pipeline?

The linting is running now. seemed the __init__.py was messing it up. a few months ago I understood init files in python3 but I couldnt tell you now why that makes it an invalid package. The tests pass and the server works

I started working on some lint fixes but just felt like i was sweeping them under the rug. `initiatives-api-lint` has the changes I made, but didnt finish all of the lint errors.

api/data_sources.py
```diff
@@ -0,0 +1,198 @@
...
+        flat_args = {}
+
+        for k,v in self.model_key_map.items():
+            if v and type(v) is str:
+                flat_args[k] = row[v]
+            elif v and type(v) is dict:
+                # assume the model_key_map contains a custom function for parsing the dataset column
+                name = next(iter(v))
+                dataset_value = row[name]
+                flat_args[k] = v[name](dataset_value) if dataset_value else None
```
jtuckerk, line 138: This was confusing and the linter agrees

jtuckerk (commented):
the roles endpoint wasn't returning any roles, I didnt dig beyond trying the events endpoint and seeing it yielded data.

I've got a few changes in airtable I'd like to push to the roles DB if we could sync a bit on that, just to be sure I'm not breaking anything here.
Then I think this is good enough

api/data_sources.py
trippersham: it's more performant than converting to a list, but agree it's less readable (which is more important in this case). Updated the PR with the more readable version.

jtuckerk (approved):
This is gonna be better later. for now we're just rolling with it

PR #35, opened 2020-11-03
Refactored nav bar + landing page and added footer and register page

ecaoile:
Note that some buttons don't work properly yet. I'm going with the current copy we have along with the figma designs. Unfortunately, I didn't have time to build account creation on the UI side. Maybe someone else can hook that up tomorow.

If anyone is copying the card layouts, the register component has slightly cleaner code than the home page.


jtuckerk (approved):
This is awesome Earl Jay! 

not necessary, but if we're not using UBI-square.png lets remove it for now.

I think it's worth going ahead and merging this to move things along, but I'm noting the things that still need work (for the whole team, not just you) just so it's in one place:

index.html:
* register to volunteer button needs the short intake form
* view our initiatives: we need an initiatives page
* our initiatives nav bar item: I like the drop down, but for simplicity let's just have that point to the initiatives page so we dont need to worry about an api call there.
* donate copy for the last text card on the index page

/register
* "register yourself as a volunteer" card copy
* register here link
* tell us more about yourself: update survey to exclude name, just use email for ID 
* copy for facebook and twitter post buttons. Add other buttons as well?
* add post request to store email addresses for manual follow ups

/initiatives - the whole page :)

trippersham:
Capturing a couple of little things wearing my QA hat but this looks fantastic, let's pull it in for now to others can build off of it!

trippersham (approved):

PR #50, opened 2020-11-06
Diff vol roles

jtuckerk:
Adding details link to a volunteer role and a role type enum.

updating the client to use the new concept of a role type choose between "apply now" and "tell me more" links.

### DIFF
--- a/client/src/components/InitiativeDetail.js
+++ b/client/src/components/InitiativeDetail.js
@@ -109,22 +109,42 @@ function InitiativeDetail(props) {
           roles.push(<h2 key="roles_header_special" className="header-2" style={{textAlign: 'center'}}>Volunteer Roles</h2>);
         }
         const role = detail["roles"][i];
-        roles.push(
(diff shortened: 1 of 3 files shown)


client/src/components/InitiativeDetail.js
```diff
@@ -109,22 +109,42 @@ function InitiativeDetail(props) {
...
+                <a href={role["details_url"]}>
+                  <Button
+                    variant="outline-info"
+                    style={{ padding: '.35rem 1.5rem' }}
+                  >
+                    Tell me more!
+                  </Button>
+                </a>
+              </div>
+            </Col>
```
jtuckerk, line 145: This is an ugly copy paste, and I didnt write tests, but this change makes me feel 10x better about pushing out the Portal  on Friday

jtuckerk:
image

trippersham (approved):

trippersham:
@jtuckerk can you fix the mypy error and then let's push it!

PR #52, opened 2020-11-06
Release v0.7

trippersham:


api/api.py
```diff
@@ -0,0 +1,82 @@
...
+    , generate_fake_volunteer_role, generate_fake_volunteer_event, generate_fake_initiative
+from data_sources import Dataset, DataSource, DataSourceType, generate_hf_mysql_db_address, DataSink
+from data_source_maps import hf_events, hf_volunteer_openings, hf_initiatives
+import logging
+
+logging.basicConfig(level=logging.INFO)
+
+app = FastAPI()
+
+origins = ['*']
```
Swimburger, line 17: I don't think it's an issue to use a wildcard, but it would be good to be able to explicitly configure the origins we want to allow through config files or environment variables that we can set per environment.

api/api.py
```diff
@@ -0,0 +1,82 @@
...
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=origins,
+    allow_credentials=True,
```
Swimburger, line 22: If we allow credentials we should not use a wildcard for origins.

trippersham: Agree we need to lock these settings down from "let anyone in," but I think the risk is small right now. Do you think it's a significant enough risk to block the release right now, or can we fix it moving forward?

Swimburger (approved):
Looks good to me. (get another pair of eyes on the Python though)

api/api.py
Swimburger: Not a blocker. I don't think we have auth yet anyway.

api/api.py
```diff
@@ -0,0 +1,82 @@
...
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+hf_db = DataSource(data_source_type=DataSourceType.MYSQL,
+        address=generate_hf_mysql_db_address('35.188.204.248','airtable_database','no_pii','humanity-forward_hf-db1-mysql_no_pii'))
```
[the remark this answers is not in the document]
trippersham: Definitely :)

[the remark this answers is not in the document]
MiniguyBrendan: Echoing Niels on this--I think setting up a simple config was one of the next prioritized tasks, and db info will definitely be part of that.

PR #55, opened 2020-11-12
Small refactor of API for dev convenience

trippersham:
* Replaced hydrating model via introspection with declarative
validator decorators in the models
* Fixed fake_data_utils
* Added stopgap fallback to generated fake data for when HF gcloud
authentication fails

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -1,8 +1,10 @@
-from pydantic import BaseModel, EmailStr
-from typing import Optional, Union, List
+from pydantic import BaseModel, EmailStr, validator
+from typing import Optional, Union, List, Type, Any
 from enum import Enum
 from uuid import UUID, uuid4
 from datetime import datetime, timezone
+
+from data_sources import get_dataset_for_model
 
 Url = str
 MarkdownText = str
@@ -37,13 +39,17 @@ class VolunteerRole(BaseModel):
     priority: Priority = Priority.LOW
     role_type: RoleType
     point_of_contact: Optional[Person]
-    num_openings: int = 1
+    num_openings: Optional[int] = 1
     min_time_commitment: Optional[int]
     max_time_commitment: Optional[int]
     overview: Optional[MarkdownText]
     benefits: Optional[MarkdownText]
     responsibilites: Optional[MarkdownText]
     qualifications: Optional[MarkdownText]
+
+    @validator('point_of_contact', pre=True)
+    def fetch_point_of_contact(cls, v):
+        return fetch_model_for_field(cls, v, Person)
 
 class VolunteerEvent(BaseModel):
     event_uuid: UUID = uuid4()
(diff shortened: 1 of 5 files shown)


api/models.py
```diff
@@ -57,18 +63,61 @@ class VolunteerEvent(BaseModel):
...
+        if v and type(v[-1]) is str:
+            # v is a list of primitive strings, attempt to fetch list of pydantic objects
+            # this filters out nested objects that fail validation
+            return [instance for key in v if (instance := get_model_for_primary_key(key, field_model))]
+        else:
+            return v
+    else:
+        # v is either a pydantic object, a valid dict of keys for that model, or an invalid input
+        # Let pydantic field validation handle organically
+        return v
```
jtuckerk, line 123: Is there a case where we'll see a pydantic object coming through this path?

I can see the object being a dictionary, and I like this flexibility.

this is fine, I'm just curious

jtuckerk (approved):
I like this a lot! thanks!

Do you think it would be possible to create a base class which auto generates the validator decorated functions using just a mapping like:
{"roles":  VolunteerRole}
or 
{"events": List[VolunteerEvent]}

Not asking you to do it, just wondering if you already went down that path and ran up against some blockers

api/models.py
trippersham: Yup, that's exactly what happens in generate_fake_initiative() in fake_data_utils

trippersham:
> I like this a lot! thanks!
> 
> Do you think it would be possible to create a base class which auto generates the validator decorated functions using just a mapping like:
> {"roles": VolunteerRole}
> or
> {"events": List[VolunteerEvent]}
> 
> Not asking you to do it, just wondering if you already went down that path and ran up against some blockers

Had the same idea (but with that base class doing it automatically via introspection without needing that mapping). I haven't explored it yet, wanted it to be declarative and clear for now! Would love to pair on that in the future.

api/models.py
```diff
@@ -57,18 +63,61 @@ class VolunteerEvent(BaseModel):
...
     initiative_external_id: str
     name: str
     details_url: Optional[Url]
     title: str
-    hero_image_url: Url = generate_placeholder_image()
+    hero_image_url: Optional[Url] = generate_placeholder_image()
     content: MarkdownText
-    roles: List[VolunteerRole] = []
-    events: List[VolunteerEvent] = []
+    roles: Optional[List[VolunteerRole]] = []
```
MiniguyBrendan, line 78: Is there a benefit to marking these Optional?

MiniguyBrendan (approved):

api/models.py
trippersham: With this, the default value, and the validator we're guaranteed to at least get an empty list. This specifically allows us to accept either the missing key or `None` as the value without failing validation (which happens for many of the initiatives coming from HF's airtable).

PR #62, opened 2020-11-20
update logo

jtuckerk:
HF-RR-long-logo -> hfrr_logo.png from ericka


trippersham (approved):
Not sure how else we can test the "production" change until we try it, so let's do it!

PR #65, opened 2020-11-21
Enable multiple environments with different databases

adamfortuna:
### Goals

- Allow for all data currently used by the volunteer portal to be accessed via a local database
- Create a setup script to create and populate a local database to enable development without production database access

### Finish relationship updates
- [x] Figure out how to create a new initiative with an array of `role`, `event`, or `initiative` objects that get saved in a single database transaction
- [x] Get the relationship working for initiatives
  - [x] Update tests to verify the model relationship, schema and API
  - [x] Fix duplicate relation bug in this test case: `ENV=test python -m pytest -k test_initiative_roles_scoped`
- [x] Get the relationship working for initiatives -> events
  - [x] update tests to verify the model relationship, schema and API
- [x] Get the relationship working for initiatives -> highlighted items
    - [x] update tests to verify the model relationship, schema and API
- [x] Add back and updated commented out tests
- [x] Get `PersonalDonationRequests` endpoint working

### Finish Postgres + MySQL Work
- [x] Abstract out all mysql/postgres specific column information into a setting that's dynamically pulled from `settings` based on the configured database engine according to `config.yaml`.

### Testing

- [x] Get testing working on GitHub

## Notes

The current approach in this branch is to move our models from Pydantic models to SQLAlchemy models. Pydantic models will still be used for our return types used in FastAPI. This is similar to the approach laid out in the [sample FastAPI codebase](https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/schemas/item.py). This will allow us to have a different Pydantic model for each API endpoint. For example, if we return a different data representation of an object in the LIST API endpoint as in the CREATE API endpoint for a model.

There are still a bunch of things to work out: validation, testing, verification that the dynamic database adapters work (that use postgres locally, and mysql bigtable on production) and more.

## Setup

Steps to get this branch working will eventually be in the readme. Until then, here's a brief intro:

Install Postgres locally. You'll need to have a `postgres` table in your local database -- which should always be created by default.

Next, make a copy of the `api/config.yaml.example` file to `api/config.yaml`. Make any changes to that copy based on your local development setup usernames, passwords and IP addresses.

With Postgres running, setup the API app like normal – either with virtualenv or conda. Install all dependencies.

run `python bootstrap.py` (or `make bootstrap`). This will create the local database, create the tables and populate them with sample data. After that you should be able to run the app like normal.

This should work for running tests and accessing the website. If you'd like to setup a different test database, you can set that up in the config.yaml file and bootstrap it using `ENV=test make bootstrap` to create it. Side note: `make test` uses whatever database you have configured in the `test` section of your config.

To run a single test case, like this one that's failing, you can run `ENV=test python -m pytest -k test_initiative_roles_scoped
`.


api/settings.py
```diff
@@ -0,0 +1,50 @@
...
+    with open("config.yaml", 'r') as f:
+        config = yaml.load(f, Loader=yaml.FullLoader)[ENV]
+    return config
+Config = read_config()
+
+
+# Generate Database URLs based on the DB connection information
+def generate_hf_mysql_db_address(connection) -> str:
+    secret_path = f'projects/humanity-forward/secrets/{db_secret_key}/versions/latest'
+    db_pass = secret_client.access_secret_version(request={"name": secret_path}).payload.data.decode('UTF-8')
```
jtuckerk, line 23: It'd be nice if this could be skipped if a password is supplied so this could work fully offline/without any hf GCP permissions.
I know this isn't done so don't worry about responding, but this probably won't be the last comment on incomplete code.

api/models/volunteer_event.py
```diff
@@ -0,0 +1,32 @@
...
+    def hero_image_url(self):
+        return self.hero_image_urls[0]['url'] if self.hero_image_urls else placeholder_image()
+
+    @hybrid_property
+    def point_of_contact(self):
+        return Person(name=self.name)
+
+    def __repr__(self):
+        return "<VolunteerEvent(event_uuid='%s', event_external_id='%s', name='%s')>" % (
+                                self.event_uuid, self.event_external_id, self.name)
```
jtuckerk: I like this organization quite a bit

api/api.py
```diff
@@ -1,82 +1,55 @@
...
 @app.get("/api/", response_model=str)
 def root() -> str:
     return "Hello from the Humanity Forward Volunteer Portal Dev Team"
 
-@app.get("/api/volunteer_roles/", response_model=List[VolunteerRole])
-def get_all_volunteer_roles(fake_data: bool = False) -> List[VolunteerRole]:
-    return roles_dataset.get_linked_model_objects() if not fake_data else generate_fake_volunteer_roles_list()
+@app.get("/api/volunteer_roles/", response_model=List[VolunteerRoleSchema])
+def get_all_volunteer_roles(db: Session = Depends(get_db)) -> List[VolunteerRoleSchema]:
+    return db.query(VolunteerRole).all()
```
jtuckerk, line 36: so the sqlalchemy base models are autoconverted to pydantic models? that's pretty neat!

I'm wondering if we could/would want to try to combine the schemas and models so we only have a single model/schema. I'm concerned about the complexity of this as a volunteer project. It might be a more complicated base class, but I'm hoping it'd simplify additions and debugging. I'm imagining having a mismatch somewhere in the schema->model->db chain could be really annoying to debug.

with this setup I'm imagining we could still have additional schemas defined for endpoints from which we only want to serve a subset of fields.

api/settings.py
jtuckerk: ah I think I misunderstood this on the first go around. There was no generator supplied in the yaml so it was breaking there but I see now I can specify the url generator to use either local or remote in the yaml

Calvinxc1:
I added a tweak that'll let us autoload Postgres for our needs using docker-compose, it's in the `/postgres` folder. One can just run `docker-compose up -d` in that folder (assuming docker-compose is [installed](https://docs.docker.com/compose/install/)) and it'll start up a containerized Postgres instance. The existing bootstrap approach works just fine for loading the test data, though we may want to look at getting a fully pre-initialized container with the test data already in it, ideally.

api/api.py
jtuckerk: > I added a tweak that'll let us autoload Postgres for our needs using docker-compose, it's in the `/postgres` folder. One can just run `docker-compose up -d` in that folder (assuming docker-compose is [installed](https://docs.docker.com/compose/install/)) and it'll start up a containerized Postgres instance. The existing bootstrap approach works just fine for loading the test data, though we may want to look at getting a fully pre-initialized container with the test data already in it, ideally.

Nice! I like moving towards dockerizing everything

How easy is it to define the test data to pre-initialize the db with? I like the set up now since, to me, editiing the python to add rows or edit the schema is intuitive, and also might means if people want to use fake data for testing theyve already written those util functions

jtuckerk:
I've rebased this change onto the development branch.

I haven't had the time I've wanted to get this fully spruced up to merge into develop. I think it is getting close to at least unblock people for more convenient local development.

The main piece that I think needs to be finished before we can merge this into develop is the tests. Im uncertain about some of the modules are set up and the dependencies referenced. I mostly have used python with bazel to manage dependencies and feel a bit out of my depth on this. @Calvinxc1 or @trippersham if either of you have opinions about how this should be done, please chime in

api/api.py
adamfortuna: > so the sqlalchemy base models are autoconverted to pydantic models? that's pretty neat!

Yep! It duck types the passed in object as whatever the schema is and converts it. I was surprised when it worked to be honest haha.

> I'm wondering if we could/would want to try to combine the schemas and models so we only have a single model/schema. I'm concerned about the complexity of this as a volunteer project. It might be a more complicated base class, but I'm hoping it'd simplify additions and debugging. I'm imagining having a mismatch somewhere in the schema->model->db chain could be really annoying to debug.

That is a concern for sure. There are some libraries that auto-convert from a model to a schema like [pydantic-sqlalchemy](https://github.com/tiangolo/pydantic-sqlalchemy). Those would convert everything, then we'd overwrite everything that's renamed for the front-end.

> with this setup I'm imagining we could still have additional schemas defined for endpoints from which we only want to serve a subset of fields.

adamfortuna:
> I added a tweak that'll let us autoload Postgres for our needs using docker-compose, it's in the `/postgres` folder. One can just run `docker-compose up -d` in that folder (assuming docker-compose is [installed](https://docs.docker.com/compose/install/)) and it'll start up a containerized Postgres instance. The existing bootstrap approach works just fine for loading the test data, though we may want to look at getting a fully pre-initialized container with the test data already in it, ideally.

Ohh nice! I'm looking forward to giving that a try.

Calvinxc1:
Ok, just merged a new docker setup. You should be able to do a `docker-compose build` followed by a `docker-compose up -d` to run the Postgres database. It'll be accessible on the same port that it is by default (5432). Also adminer is running at port 8080 so you have a GUI interface to explore the database. Enjoy!

Calvinxc1:
Getting the following when I try to run `docker-compose build` on the new code:
```
Step 6/6 : RUN npm run build
 ---> Running in e17d1234133e

> volunteer-portal@0.5.0 build /srv/client
> react-scripts build

Creating an optimized production build...
Failed to compile.

./src/styles/base.scss
Error: Missing binding /srv/client/node_modules/node-sass/vendor/linux-x64-72/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 12.x

Found bindings for the following environments:
  - Linux 64-bit with Node.js 14.x

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to download the binding for your current environment.


npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! volunteer-portal@0.5.0 build: `react-scripts build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the volunteer-portal@0.5.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-12-22T14_40_40_311Z-debug.log
```

adamfortuna:
> Getting the following when I try to run `docker-compose build` on the new code:

Hrm, that's a weird one. I just tried clearing all of my docker containers and running a `docker-compose build` from scratch and it's not happening. What OS are you using? I'm wondering if this is a Mac vs Windows issue. Reading through [this article](https://jdlm.info/articles/2019/09/06/lessons-building-node-app-docker.html) on Docker makes me think there's going to be some additional fixes we'll need to add to Windows if that's the case in order to map users. Hopefully it's something more simple though.

Calvinxc1:
> > Getting the following when I try to run `docker-compose build` on the new code:
> 
> Hrm, that's a weird one. I just tried clearing all of my docker containers and running a `docker-compose build` from scratch and it's not happening. What OS are you using? I'm wondering if this is a Mac vs Windows issue. Reading through [this article](https://jdlm.info/articles/2019/09/06/lessons-building-node-app-docker.html) on Docker makes me think there's going to be some additional fixes we'll need to add to Windows if that's the case in order to map users. Hopefully it's something more simple though.

Linux ;-)

adamfortuna:
> Linux ;-)

Back to the drawing board. 😅

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+    depends_on:
+    - db
+    ports:
+    - 8081:8081
+    volumes:
+    - ./api:/api
+
+  client:
+    build: ./client
+    command: npm run start
```
jtuckerk: Let's do "npm run client"
run start runs from the build directory and requires a rebuild to pull in changes.
"npm run client" will allow changes to be hot reloaded.

When we break out services/compose files for prod vs dev we'll use npm run start in prod there to use the optimized build

Makefile
```diff
@@ -0,0 +1,30 @@
...
+shell-api:
+	docker-compose run --rm api bash
+test:
+	docker-compose run --rm api python -m pytest tests/
+validate:
+	docker-compose run --rm api mypy /api
+db-reload-dev:
+	docker-compose run --rm api python bootstrap.py
+db-reload-dev:
+	docker-compose run --rm api ENV=test python bootstrap.py
```
jtuckerk: This wasn't working for me

api/data_sources.py
```diff
@@ -173,14 +174,47 @@ def create_linked_model_object(self, row: ResultProxy) -> Optional[Any]:
...
+        existing_linked_datasets = {k[2]:v for k,v in datasets.items() if k[2]}
+        args = {}
 
-def get_dataset_for_model(model: Type) -> Optional[Dataset]:
-    existing_linked_datasets = {k[2]:v for k,v in datasets.items() if k[2]}
-    if model in existing_linked_datasets.keys():
-        return existing_linked_datasets[model]
-    else:
-        return None
+        # iterate over the parameters of the model class
```
jtuckerk: I think this whole file can be deleted, but this section in particular looks like a regression to an older commit

api/deploy-prod-app.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 branch_name=$(git symbolic-ref --short -q HEAD)
-if [ "$branch_name" = "production" ]; then
+if [ "$branch_name" = "master" ]; then
```
jtuckerk: this shouldn't be reverted to master

client/deploy-prod-client.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 branch_name=$(git symbolic-ref --short -q HEAD)
-if [ "$branch_name" = "production" ]; then
+if [ "$branch_name" = "master" ]; then
```
jtuckerk: This shouldn't be reverted to master

jtuckerk (changes requested):
I think we're pretty close to being good enough to merge this with develop.

It looks like this PR reverts some changes. I think was happened was a rebase/merge/gitword in  6a4d9bb back onto master. (edit: master is gone, i'm not sure exactly what happened but it was definitely caused by switching from master to develop after this branch was started) the should have been a rebase onto develop

2 PRs to merge to this one:
#76 makes slight changes to the docker configs that I think will same others some frustration
#75 is working, but not thoroughly tested, code for the airtable data pipelines

Some other things that I think we can finish later:

- the data source/sink code is now obsolete, I believe
- editing the db columns will require changes in the models, schemas, pipeline converters, and fake data utils. while this isnt ideal, I dont think it's a blocker.
- linting and type annotations (I need to do this in the pipeline code as well)

docker-compose.yml
adamfortuna: Sounds like a plan, changing this one.

Makefile
adamfortuna: Updated these to use a new test container and updated the makefile.

api/data_sources.py
adamfortuna: Yeah, I think it can. If we need parts of it, we can move them into the new spots later. Deleting these for now.

api/deploy-prod-app.sh
adamfortuna: Good catch!

client/deploy-prod-client.sh
adamfortuna: Good catch!

api/api.py
```diff
@@ -56,7 +56,7 @@ def get_initiative_by_external_id(initiative_external_id, db: Session = Depends(
     return db.query(Initiative).filter_by(initiative_external_id=initiative_external_id).first()
 
 @app.post("/api/donation_link_requests/", response_model=DonationEmailSchema)
-def create_donation_link_request(email: str, db: Session = Depends(get_db)) -> Optional[DonationEmailSchema]:
-    donationEmail = DonationEmail(email=email)
-    DonationEmail.insert(donationEmail)
+def create_donation_link_request(donationEmail: DonationEmailSchema, db: Session = Depends(get_db)) -> DonationEmailSchema:
+    db.add(DonationEmail(email=donationEmail.email))
```
jtuckerk, line 60: The old way looked weird. I was surprised it worked. or did it not?

api/schemas/donation_email.py
```diff
@@ -1,13 +1,31 @@
...
+    )
+    return cls
+
+@form_body
 class DonationEmailSchema(BaseModel):
-    donation_uuid: UUID
     email: str
-    request_sent: datetime
+
+    @validator('email')
```
jtuckerk: Theres a pydantic type we can use to do this validation
`from pydantic import EmailStr`

client/src/App.js
```diff
@@ -1,11 +1,9 @@
 import React from 'react';
```
jtuckerk: I believe this was an unintended revert too, along with all of the other changes in client/src I think

jtuckerk (changes requested):
I think there are still a bunch of changes in the client that revert a fair bit of work in development

should there be any changes in this PR to the files in client/src? 

The rest of the changes look good tho!

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+    - ./client:/client
+    - client_node_modules:/client/node_modules
+
+  # Proxy allows the client app to access the API on the same host
+  # This avoids cross-domain issues.
+  # This is the main entrypoint to the application.
+  proxy:
+    image: nginx
+    depends_on:
+    - api
```
jtuckerk, line 50: can we add api-test here? IIUC this only affects the spin-up order but wont spin up both api and api-test at the same time when the proxy reloads

jtuckerk: Nvm, with the PR i have  out this may be obsolete

jtuckerk:
changes from #75, #78 and #80 all merged into and on top of this and appear to be working in branch revert+docker+db_pipeline

api/api.py
adamfortuna: I haven't been able to validate the previous way works, but I know the new way works at least. I wanted to make it cleaner, but gave up eventually and went the less-pretty way.

client/src/App.js
adamfortuna: Off, yeah there shouldn't be any changes in Client other than cleaning up an API request path or two to no longer include a trailing slash. Must've had conflicts during a merage. I'll replace that.

jtuckerk (approved):

PR #66, opened 2020-11-24
added nav item to go back to main HF website

ecaoile:


jtuckerk (commented):
technically LGTM. I have thoughts about the button text, but I'm not sure what it should be.

I want it to be shorter, and many people may not come from the main site but "Main HF Site" seems worse for some reason.

I really wish HF got humanityforward.com without the "move"

@trippersham this is up on staging at https://vol-dev.movehumanityforward.com/

image

image

image

image

trippersham:
How do we feel about "Return to Parent Site"? I don't think there's any way to include "Humanity Forward" in the text of the button since it will be too long and my preference (unsurprisingly) is to make it more generic.

I was also tossing around "Organization's Main Site" but I do like the "Return" or "Go Back" language because it makes it clear to the user that they're leaving the portal

trippersham:
Also any chance we can center those items in the mobile drawer? That horizontal alignment keeps bugging me

ecaoile:
> 
> 
> Also any chance we can center those items in the mobile drawer? That horizontal alignment keeps bugging me

I'm not sure why I added that margin in the first place, but I've gotten rid of it along with changing the text. How does it look now?
image

jtuckerk (approved):

trippersham (approved):

trippersham:
Looks great!

PR #67, opened 2020-11-25
Update style

stran335:
Minor updates to css to exactly match HF Style guide.


jtuckerk (changes requested):
It looks like you had the unfortunate luck of having started this change on a branch from before a big change we made to the branching strategy.

There may be a git-savvy way to fix this, but I would just check out a fresh branch from the develop branch and copy your changes from client/src/styles/_variables.scss and client/src/styles/header.scss (I think those are the only 2 files you intend to change, right?)

That should fix the merge conflicts

stran335:
No problem! Yes I changed just a couple lines in those two files. This has been more a an exercise in me trying to learn how git works ha.  I'll resync with develop.

PR #68, opened 2020-11-29
update css to exactly match HF style guide

stran335:
Update css to exactly match HF style guide colors.
Link, hover, background white, black text


jtuckerk (approved):
Thanks!

PR #69, opened 2020-12-05
collapse nav menu when clicking to register

stran335:
On mobile - ensure nav menu collapses after clicking "Register to Volunteer" button link.


Swimburger:
This works, though I'd reuse the existing `() => setTimeout(() => setExpanded(false), 100)`.
This is not really a scenario that users would run into, but if you click on the button on desktop viewport size and then make your viewport smaller, the navigation menu is expanded instead of collapsed.

Can we introduce a local function `collapse` to hold `() => setTimeout(() => setExpanded(false), 100)`. Then you can set onClick on all instances like `onClick={collapse}`.

Swimburger (approved):

PR #71, opened 2020-12-14
Authentication Foundation & Secrets Management

trippersham:
Skeleton in place for OAuth2 and cookie-based JWT Auth strategy with working login and logout via Google and Github. Secrets are now managed through Google Secret Manager correctly and used for local dev, in CI, and (theoretically automatically) in deployed environments. Updated config.yaml and Makefile for convenience and clarity. Removed some cruft. Added docs.


api/api.py
```diff
@@ -31,6 +49,31 @@ def get_db():
...
+    redirect_uri = request.url_for('auth')
+    return await oauth.google.authorize_redirect(request, redirect_uri)
+
+@app.get("/api/auth")
+async def auth(request: Request):
+    try:
+        token = await oauth.google.authorize_access_token(request)
+    except OAuthError as error:
+        return HTMLResponse(f'<h1>{error.error}</h1>')
+    user = await oauth.google.parse_id_token(request, token)
```
adamfortuna: This is super concise - I like it! Are you thinking we'd also have a `user` model in our database that's created/loaded here based on the auth info? A user has many authentications kind of setup? This might already be happening within the library too.

.gitignore
```diff
@@ -120,6 +120,7 @@ dist
 *.pyc
 *.mypy
 
+api/config.yaml
```
jtuckerk: I think we should just keep config.yaml in git. Since we won't be putting password directly in any config files (using google secret manager instead) I think it'll just save a step of having everyone copy over the example config.

jtuckerk: Here's an example of how the secret manager is used for the DB password. https://github.com/HF-RapidResponse/volunteer-portal/blob/develop/api/settings.py#L21

We can store the client ID and secret in separate 'secrets' and have a pair for prod and a pair for dev and just use all the same code paths, or if we want the dev version to work without internet/GCP project permissions set up (Slight preference because of this) we can have it toggle-able from the config like Adam did with the db_url/secret generators

api/Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY: test
```
jtuckerk: Can we delete this file? and move anything not covered by the current makefile at the project root have docker version of anything that may be missing if necessary

api/data_sources.py
```diff
@@ -0,0 +1,180 @@
+from pydantic import BaseModel
```
jtuckerk: Can we keep this file deleted?

api/models/personal_donation_link_request.py
```diff
@@ -0,0 +1,13 @@
+from models.base import Base
```
jtuckerk: Keep deleted as well?

api/requirements.txt
```diff
@@ -8,8 +8,13 @@ uvicorn[standard]
 gunicorn
 requests
 sqlalchemy
+sqlalchemy-utils
```
jtuckerk: I think (not positive) this is no longer needed)

api/schemas/personal_donation_link_request.py
```diff
@@ -0,0 +1,13 @@
+from pydantic import BaseModel, EmailStr
```
jtuckerk: delete this file?

api/settings.py
```diff
@@ -38,7 +38,7 @@ def create_connections(generators, config):
         connections[key] = config[key]
         connections[key]['url'] = generators[config[key]['generator']](config[key])
     return connections
-Connections = create_connections(db_url_generators, Config)
+Connections = create_connections(db_url_generators, Config['databases'])
```
jtuckerk, line 69: with the updated config setup only a single DB is expected so this has been flattened and I think this is just a missed fast forward/unwanted reversion

api/tests/todo_test_data_sources.py
```diff
@@ -0,0 +1,101 @@
+import pytest
```
jtuckerk: delete this file?

client/package.json
```diff
@@ -49,6 +49,5 @@
   "devDependencies": {
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.5"
-  },
-  "proxy": "http://localhost:8000"
+  }
```
jtuckerk, line 52: TODO for me: Check this doesnt break things once I'm set up with IDs and secrets

docs/auth_security.md
```diff
@@ -0,0 +1,26 @@
+# Auth & Security
```
jtuckerk, line 1: docs 
image

api/auth.py
```diff
@@ -0,0 +1,110 @@
...
+        authorize_params=None,
+        api_base_url='https://api.github.com/',
+        client_kwargs={'scope': 'user:email'},
+)
+
+@router.get("/profile")
+def get_profile(Authorize: AuthJWT = Depends()) -> Dict:
+    Authorize.jwt_required()
+
+    current_user = Authorize.get_jwt_subject()
```
jtuckerk, line 56: I'm struggling to find the docs on this. Will this be a auth client specific identifier?

jtuckerk (changes requested):
Looks good! I haven't poked around yet and want to do that before final approval but wanted to get the client IDs and secrets set up with secret manager first if possible. Do you have keys from the HF GCP project and the github org or did you use your own for testing? 

I can work with you to get those created for test and dev. I'm also wondering if we don't need them for dev and can set up a way to do auth locally that we'd use for development and unit tests.

I think there are a few files/changes that got reverted/weren't fast forwarded to the current state of dev. I've added comments inline for those

api/settings.py
trippersham: This is actually for the new config.yaml structure that pulls the DB settings into a 'databases' key. However please check to make sure what lives under that subkey is what you're expecting with the new flattened structure within it.

api/auth.py
trippersham: [Here](https://indominusbyte.github.io/fastapi-jwt-auth/) are the docs for this: I will include them in our auth security docs

Makefile
```diff
@@ -8,12 +11,18 @@ up:
...
 
 # Api
 shell-api:
 	docker-compose run --rm api bash
 test:
+	@echo "${ORANGE}WARNING: this rebuilds the api and db images to use the test configuration\n\
+	You will need to run \`make recreate-api; make recreate-db; make db-reload-dev\` \
+	to restore your local development environment \n\n${NC}"
+	
+	docker-compose $(TEST) build --no-cache api db
```
jtuckerk: does the auth stuff need to be rebuilt to be reconfigured? rebuilding the api is a bummer since its so slow :(

jtuckerk: actually on second thought, I think a "down" and "up" should work even if it does have to restart. I've tested it by removing this build line and running:
`docker-compose up`, then `make test` and saw that it failed, then ran `docker-compose down` followed by `make test`. so I think we could replace that line with just docker-compose down and hopefully save a lot of waiting, especially when people are running make test multiple times in a row

and then with respect to the above warning, just a docker-compose down and docker-compose up should bring back the full development environment.

api/auth.py
jtuckerk: thanks

api/settings.py
jtuckerk: the  structure within looks good

docs/dependencies.md
```diff
@@ -47,3 +47,13 @@ volunteer-portal (develop) $ git commit -am "Added momentjs library"
 ```
 
 Running `npm install --save momentjs` within the docker container updated the `package.json` and the `package-lock.json` in the client directory – just like it would if you were working locally. After that you can commit, push and work just like you would if this application ran using your local node environment.
+
+## Recreating Docker Images
+
+Occassionally the docker images do not automatically update when new dependencies are added (e.g. when switching branches). If you are receiving errors for a dependency that's already declared in a dependencies file, try rebuilding the corresponding service using the following commands:
+* `make recreate-api`
```
jtuckerk, line 54: same comment from above. i think docker-compose down followed by an up should fix all of this. please let me know if I'm missing something/that only works in simple cases!

jtuckerk (approved):
looks good, just some notes on your docker commands. The make test command will take multiple minutes so I want to be really sure that build step is necessary before we check it in

PR #75, opened 2021-01-09
Tucker airtable data pipeline

jtuckerk:
Creating a airtable table syncing pipeline.
So far this is only set up for events

todo:

- [x] initiatives
- [x] roles
- [x] more testing


api/api.py
```diff
@@ -37,26 +38,45 @@ def get_all_volunteer_roles(db: Session = Depends(get_db)) -> List[VolunteerRole
...
-    return db.query(Initiative).filter_by(initiative_external_id=initiative_external_id).first()
+    return db.query(Initiative).filter_by(external_id=initiative_external_id).first()
 
 @app.post("/api/donation_link_requests/", response_model=DonationEmailSchema)
 def create_donation_link_request(email: str, db: Session = Depends(get_db)) -> Optional[DonationEmailSchema]:
     donationEmail = DonationEmail(email=email)
     DonationEmail.insert(donationEmail)
     return donationEmail
+
+def SyncRunner(db, sync_fn) -> str:
```
adamfortuna: What would you think about moving from here down to a different file? I saw @trippersham did something like that in the auth branch, ( https://github.com/HF-RapidResponse/volunteer-portal/blob/auth_poc/api/api.py#L11 ) which I think would be a good way to add additional routes without putting everything in api.py.

api/db_sync_pipelines/pipelines.py
```diff
@@ -0,0 +1,34 @@
+from db_sync_pipelines.airtable_sync import RunAirtableSync, AirtableLoader
+from db_sync_pipelines.airtable_event_converter import AirtableEventConverter
+from db_sync_pipelines.airtable_role_converter import AirtableRoleConverter
+from db_sync_pipelines.airtable_initiative_converter import AirtableInitiativeConverter
+
+def RunEventsSync(db):
+  table_name = "[Events]"
+  table_key = 'appXC8UaNIuMXpnGB'
```
adamfortuna, line 8: Do you think we could split these out into a settings variable? It'd be nice to be able to test this using a copy of the Airtable without having access to production.

api/models/volunteer_role.py
```diff
@@ -12,31 +12,31 @@
...
+    priority = Column('priority', Enum(Priority))
+    team = Column('team', ARRAY(String(255)))
+    team_lead_ids = Column('team_lead_ids', ARRAY(String(255)))
     num_openings = Column(Integer, default=1)
     min_time_commitment = Column('minimum_time_commitment_per_week_hours', Integer)
     max_time_commitment = Column('maximum_time_commitment_per_week_hours', Integer)
     overview = Column('job_overview', Text)
     benefits = Column('what_youll_learn', Text)
-    responsibilites = Column('responsibilities_and_duties', Text)
+    responsibilities = Column('responsibilities_and_duties', Text)
```
adamfortuna, line 30: lol, good catch!

api/models/initiative.py
```diff
@@ -12,58 +12,61 @@
...
+    external_id = Column('id', String(255))
+    initiative_name = Column('initiative_name', String(255))
+    order = Column('order', Integer)
     details_url = Column('details_link', String(255), nullable=True)
     hero_image_urls = Column('hero_image_urls', ARRAY(JSON))
     content = Column('description', String(255))
     role_ids = Column('roles', ARRAY(String))
     event_ids = Column('events', ARRAY(String))
+    airtable_last_modified = Column('airtable_last_modified', DateTime, nullable=False)
+    db_last_modified = Column('db_last_modified', DateTime, nullable=False)
```
adamfortuna: This is 100% personal preference, but I always call this column `updated_at`. It looks like there's a way to automatically update this column too: https://stackoverflow.com/questions/13370317/sqlalchemy-default-datetime

```
from sqlalchemy.sql import func
updated_at = Column('updated_at', DateTime, onupdate=func.now())
```

Could be one wy to slim down a little code.
With that, SQLAlchemy will automatically update that column's value whenever it changes.

api/tests/db_sync_pipelines/events_example_response.json
```diff
@@ -0,0 +1,223 @@
...
+      "Event / Action Type": "Humanity CALLS phone bank",
+      "Priority": "High",
+      "Vols. Recruited": 69,
+      "PTG (Recruitment)": {
+        "specialValue": "Infinity"
+      },
+      "Last Modified": "2020-11-20T16:42:54.000Z",
+      "Created by": {
+        "id": "usrg7iGBoicYG7cfL",
+        "email": "tom@tomjeffreyinnovations.com",
```
adamfortuna: Might want to remove these emails, and change some names to use `@example.com` if they're real ones.

api/tests/db_sync_pipelines/events_example_response.json
```diff
@@ -0,0 +1,223 @@
...
+      ],
+      "Priority": "Low",
+      "Vols. Recruited": 18,
+      "PTG (Recruitment)": {
+        "specialValue": "Infinity"
+      },
+      "Last Modified": "2020-11-20T16:42:54.000Z",
+      "Created by": {
+        "id": "usrg7iGBoicYG7cfL",
+        "email": "tom@tomjeffreyinnovations.com",
```
adamfortuna: Might want to remove these emails, and change some names to use `@example.com` if they're real ones.

adamfortuna (changes requested):
Few minor recommendations, but looks good to me!

api/db_sync_pipelines/pipelines.py
jtuckerk: I do think it's something we should look into, but maybe further down the road.
I'm not sure I want to spin up a test version of the airtable. I don't see much benefit since any changes to the prod tables wouldn't be caught by those tests. (the airtable team is looking into how to prevent changes to the tables and columns we care about.)

It doesn't look like there are very fine grained permissions within airtable so giving permissions to a view or even a "synced" version of the table would open up all of airtable ~ish. ish: If we can change those table keys ^ we could keep the new ones private (should have kept them private to start) and only expose the global key (which we'd have to verify doesn't grant access to downloading individual table keys, or anything else we dont want)

option I like the most would be while syncing, dump a subset the json response to a cloud storage file and use something like the FakeAirtableLoader ive used in the tests to serve as a dev/test replacement for airtable

api/tests/db_sync_pipelines/events_example_response.json
jtuckerk: ah good catch!

api/tests/db_sync_pipelines/events_example_response.json
jtuckerk: thanks!

api/models/initiative.py
jtuckerk: oh yeah much better. thanks!

api/api.py
jtuckerk: done!

api/tests/db_sync_pipelines/events_example_response.json
jtuckerk: I'll rebase this and force push the changes before we merge this just to remove that from the commit graph, but wanted to leave you with the diffs for now.

api/models/volunteer_role.py
jtuckerk: :)

api/db_sync_pipelines/airtable_event_converter.py
```diff
@@ -0,0 +1,26 @@
...
+    "hero_image_urls": "fields.Event Graphic(s)",
+    "signup_url": "fields.Sign-Up Link",
+    "start_datetime": "fields.Start",
+    "end_datetime": "fields.End",
+    "description": "fields.Description",
+    "airtable_last_modified": "fields.Last Modified",
+  }
+
+  custom_transforms = {
+      'start_datetime': ParseTimestamp,
```
trippersham: I like that these are defined functions (as opposed to the lambdas I was using previously). I feel like it will be good to have a library of these common "custom" transforms when we have other data sources in the pilot even if we need to support truly one-off transforms in the future.

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+        return self.airtable_client.get_all(view=self.view)
+
+def RunAirtableSync(airtable_loader,
+                    db,
+                    response_converter,
+                    airtable_view=None,
+                    hard_delete=False):
+
+    db_model = response_converter.GetDBModel()
+    all_records = list(GetAndConvertAirtableRecords(
```
trippersham: GetAndConvertAirtableRecords is a generator but you immediately convert it to a list to then use in a loop. Unless I misunderstand generators (very possible) you're losing the performance benefits of using the generator in the first place; why not just pass the generator straight into the loop?

trippersham: Digging through the calls: why are you using a generator? There's only one network call and you get a list of all of the airtable rows (I'm assuming the memory usage of that list is not substantially different than the converted rows).

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+
+def GetAndConvertAirtableRecords(airtable_loader,
+                                 response_converter,
+                                 view=None):
+    for r in airtable_loader.GetTable():
+        yield response_converter.Convert(r)
+
+
+def GetExistingRecordIds(db, db_model, hard_delete=False):
+    query = db.query(db_model.external_id)
```
trippersham, line 77: This creates a hard requirement that all external data sources need to have a persistent unique ID associated with each row. Not sure if that's a good or bad requirement, but something we should keep in mind for the future.

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+    if not hard_delete:
+        query = query.filter(db_model.is_deleted == False)
+
+    ids = set([x.external_id for x in query.all()])
+    return ids
+
+
+def GetLatestSyncRun(db, db_model):
+    query = db.query(db_model.updated_at).order_by(db_model.updated_at.desc())
+    result = query.first()
```
trippersham, line 87: This assumes those records can only be modified by the sync pipeline. We don't have a way to edit these records from the UI today but we probably will in the future (even if we don't allow you to edit/override this data, we may have other fields like "users_with_this_role" that are updated on our side). It may be safer/more future-proofed to keep a separate updated column that only pertains to the sync pipeline.

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+    existing_record_ids = GetExistingRecordIds(db, db_model, hard_delete)
+    latest_sync_run = GetLatestSyncRun(db, db_model)
+
+    new_records = []
+    updated_records = []
+
+    for record in all_records:
+        if record.external_id not in existing_record_ids:
+            new_records.append(record)
+        elif NeedsUpdate(record, latest_sync_run):
```
trippersham, line 41: Smart, I like this

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+            updated_records.append(record)
+
+    airtable_records = set([record.external_id for record in all_records])
```
trippersham: slight suggestion to change these variable names to "_ids" so new folks don't think the contents of "new_records" and "deleted_records" are the same types

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+        logging.warning('Failed to execute %s queries: %s' % (len(db_records), e))
+        raise
+
+
+def UpdateRecords(local_records, db, db_model):
+    try:
+        for local_record in local_records:
+            # Feels like there should be a better way to do this.
+            db.query(db_model).filter(
+                db_model.external_id == local_record.external_id).delete()
```
trippersham, line 128: If/when we have other data stored on those records that doesn't come from this external source, this will be overzealous. Fine for now, but we should probably do an upsert as opposed to a delete and replace.

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+            db.add(local_record)
+        db.commit()
+    except Exception as e:
+        logging.warning('Failed to execute queries: %s' % e)
+        raise
+
+
+def UpdateRecordsAsDeleted(to_delete, db, db_model):
```
trippersham, line 136: Love a good soft delete

api/db_sync_pipelines/airtable_sync.py
```diff
@@ -0,0 +1,166 @@
...
+  return datetime.now(tz=timezone.utc)
+
+
+def ParseTimestamp(timestamp):
+  if not timestamp:
+      return None
+  return datetime.strptime(timestamp, AIRTABLE_DATETIME_FORMAT)
+
+
+class AirtableLoader:
```
trippersham, line 11: What's the rationale for having these functions live outside of the class? Making testing easier so we don't need to use a mock?

api/models/initiative.py
```diff
@@ -2,68 +2,72 @@
...
-from sqlalchemy import func, select, Column, String, Integer
+from sqlalchemy import func, select, Column, String, Integer, DateTime, Boolean, Text
 from sqlalchemy.dialects.postgresql import ARRAY, JSON, UUID
 from sqlalchemy.orm import backref, column_property, relationship, synonym
 from sqlalchemy.ext.hybrid import hybrid_property
 from pydantic import validator
+from sqlalchemy.sql import func
 from uuid import uuid4
 
 class Initiative(Base):
```
trippersham: What do you think about creating a subclass of Base that handles things like "external_id", "is_deleted", and "updated_at" (esp. depending on your thoughts on my comment on GetLatestSyncRun)? Will make it harder to accidentally forget those fields that are primarily used by the sync pipeline.

api/db_sync_pipelines/airtable_event_converter.py
jtuckerk: Yeah, I thought about pulling these out fully to a separate utils file, but since theyre so simple and the only real meat of the function is an airtable date format,  I left them in an airtable file.

jtuckerk: I'm leaning towards leaving it for now but if you feel strongly about breaking it out now, let me know

api/db_sync_pipelines/airtable_sync.py
jtuckerk: at some point in my career someone who cares a lot more about these sorts of things than I do pointed out that if a function doesnt operate on an instance's state it shouldn't be defined within the class. It avoids having to use self in the definition and the call which i guess is cleaner.
~time passes and other comments got me thinking~
I've made these part of the APIResponseConverter file, but still not a part of the class for the same reason

api/db_sync_pipelines/airtable_sync.py
jtuckerk: I probably put that in for debugging and accidentally left it. It's now removed.
I was just using a generator more for style and simplicity to avoid declaring a list and appending. Also possible I borrowed code from something that did some paging.
I'm now using list comprehension which may be bad style, idk.

api/db_sync_pipelines/airtable_sync.py
jtuckerk: May be overkill for some of these smaller tables, but if we ever decide to pull all the volunteer records in, this gives a big speedup.

api/db_sync_pipelines/airtable_sync.py
jtuckerk: good call

api/db_sync_pipelines/airtable_event_converter.py
jtuckerk: moved these to the api_response_converter file since the now timestamp is generic and airtable uses the ISO standard  (also added this and other comments to a review to avoid individual notifications for each comment I make, sorry if it formats weird and any previous annoyance)

api/db_sync_pipelines/airtable_sync.py
jtuckerk: For the data centralization work I was doing I was creating a separate syncing pipeline for each platform. definitely ended up being a lot of code reused, but then you dont have to worry as much about things like that. Since this is just the airtable syncer and it seems like we're moving away from the possibility of using more external platforms in the near to mid term I'm not too concerned
good to keep in mind for sure tho

api/db_sync_pipelines/airtable_sync.py
jtuckerk: hmmm good point.
How do you feel about leaving it as is and just moving off airtable once we have a portal UI for updating them? The idea of having 2 sources for 1 db doesn't seem fun. If we do leave this field to be updated by the UI or the pipeline then it would effectively give the UI priority since anything changed with the UI later would cause earlier Airtable changes to be ignored. which I think would be fine behavior, tho again, not excited about dealing with figuring that out completely. 20% chance I'm ignoring other good reasons to change it and am just coming up with reasons not to do any more work.

api/db_sync_pipelines/airtable_sync.py
jtuckerk: yeah,  I didnt like this. I banged my head on it for a while and didnt feel like I was getting the type of solution I thought should exist. I may not have searched "upsert" hmmm. but either way I'm still a bit wary of having multiple sources for a single record

api/models/initiative.py
jtuckerk: ooh yeah

jtuckerk: done

jtuckerk (commented):
great feedback thank you!

trippersham (approved):

PR #78, opened 2021-01-15
Set up separate services within docker

jtuckerk:
This also fixes the db file loading so that local changes are picked
  up by the DB from the appropriate env's file when it is restarted

going off recommendations here: https://docs.docker.com/compose/extends/


jtuckerk:
@adamfortuna lmk what you think of this organization!

Makefile
```diff
@@ -1,21 +1,26 @@
 .PHONY: up
 
+# Dev is the default config
+TEST = -f docker-compose.yml -f docker-compose.test.yml
+PROD = -f docker-compose.yml -f docker-compose.prod.yml
```
adamfortuna, line 5: Ohh this is neat! I didn't know you could build different envs in this way.

PR #79, opened 2021-01-21
Social share image maybe fix

ecaoile:
@trippersham I've changed the content attributes to external URLs. I'm thinking since we're sharing the prod URL, this fix won't really work until we've deployed to prod. Really wish we could test with localhost. The image will still be available in our repo. I've moved it up from being in a pics folder to being directly under the public folder.


jtuckerk:
Should we make this a hotfix? since I assume we  dont want all of the other dev stuff out in prod yet

trippersham:
Yeah, that makes sense to me

ecaoile:
Uhhh...I forgot how those work. Do we just close this PR then?

trippersham:
We at least need a PR against production, but honestly this is fine to keep open and merge too because we'll need to get it back into develop anyway. @ecaoile could you open one against production as well?

trippersham:
Also @ecaoile are we figuring out the relative paths for static images, or just using the hosted link?

ecaoile:
I changed it to external link just to be safe. Do we run any risk of losing that external host?

trippersham:
Good question. I think safe to assume that's not a risk right now, but we
can check with Caleb for the future.

jtuckerk:
+1, I think the best way is just 2 PRs off this branch.

jtuckerk (approved):
Looks good to me! I had a few hiccups building it but i think thats cuz docker had taken over the node_modules, but it all seems to be working now.


 social share og images:
image
image

trippersham (approved):

PR #84, opened 2021-02-11
Service account key doc update.

jtuckerk:
including the standard-tech-volunteer-robot service account key in the docs. if this worries you, dont fret!

This key has the same level of access of any NDA'd volunteer and only NDA'd volunteers can access it. i've vomitted my thoughts about this all over tripps DMs in talking about his latest PR. I'd be happy to lay this out a bit more clearly at this weeks meeting

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,11 @@ To run this code locally, it must first be cloned from GitHub. This can be done 
 
 The URL for cloning comes from the top right portion of the repo page.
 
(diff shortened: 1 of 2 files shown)


ecaoile (approved):

ecaoile:
Test is failing GCP credentials. Oof. I think the test needs to be updated.

jtuckerk:
> Test is failing GCP credentials. Oof. I think the test needs to be updated.

I forgot to update the key in github. new key is now in and working

PR #85, opened 2021-02-12
Release docker setup

jtuckerk:
I went around on this for a while and finally landed on this as the best solution.
The main reason being that it allows us to deploy the portal on a single machine, and can be used on cloud run/cloud build and is all very automatable. 

other options were cloning the git repo into a vm and using docker-compose. I dont like having general purpose VMs spun with the kind of authority needed to run the portal. also docker compose runs 3 separate containers and thats nearly 3x the memory we need.

Setting up an image per service would have been the simplest, but would require more networking work to configure the connection between the backend and fronted.

TODO:

- [x] remove/simplify some of the prod docker-compose stuff
- [x] clean up app engine production stuff
- [ ] ~set this up to be easier to split out to multiple instances~ It's in a fine place wrt this now

@trippersham if you could, I'd like verifying that your auth and notification work works here and input on how we should set up the configs. For now I've just copied the development templates into the staging config

pushing the latest commit has staging up and running [here](https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app), the cloud run config can be inspected [here](https://console.cloud.google.com/run/detail/us-central1/staging-volunteer-portal/metrics?project=humanity-forward) the db is working, but not all of the data is populated yet, some things are failing to sync. I've got PR #100  ready to clean that up.


trippersham:
I like running it all on one box for now for simplicity, cost, and memory-usage. We don't have major performance concerns (now that we're serving directly from our DB) or compute-intensive jobs yet. I'm curious more than worried about the airtable syncing, but that can be scheduled at time with very low site usage given we're only in a few timezones.

The approach I'm concerned about is routing everything through one URL/port. It runs counter to how development is done in industry (always a dedicated web url and a dedicated API url) and it's causing us some challenges with security. We've done it for convenience to date, but I think we should consider going to dedicated URLs for each. The nice thing is that with nginx, paired with DNS routing, we should be able to run off one box while appearing to the clients to be at separate dedicated endpoints. Will also make an eventual transition to a dedicated separate API server (that I'll propose as likely, even if we delay it for as long as we can for cost) much easier.

jtuckerk:
@trippersham definitely agree wee can split it out as soon as possible not to avoid any security issues and stick with a more conventional setup. 

I'm happy to defer to you on the approach we take as to when where how we make that change

jtuckerk:
This is ready to be reviewed. pushing the latest commit has staging up and running [here](https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app), the cloud run config can be inspected [here](https://console.cloud.google.com/run/detail/us-central1/staging-volunteer-portal/metrics?project=humanity-forward) the db is working, but not all of the data is populated yet, some things are failing to sync. I've got PR #100  ready to clean that up.

jtuckerk:
This is ready to be reviewed. pushing the latest commit has staging up and running [here](https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app), the cloud run config can be inspected [here](https://console.cloud.google.com/run/detail/us-central1/staging-volunteer-portal/metrics?project=humanity-forward) the db is working, but not all of the data is populated yet, some things are failing to sync. I've got PR #100  ready to clean that up.

trippersham (approved):

PR #87, opened 2021-02-19
Added notifications manager with base support for sending emails

trippersham:


api/notifications_manager.py
```diff
@@ -0,0 +1,53 @@
...
+            subject='This is a test Vol Portal Email',
+            html_content=f'<strong>Test email with message</strong> {notification.message}'
+        )
+
+        response = email_client.send(email)
+        if response.ok:
+            notification.status = NotificationStatus.SENT
+            notification.sent_date = datetime.utcnow()
+        else:
+            notification.status = NotificationStatus.FAILE
```
ecaoile: Is the property really "FAILE" and not "FAIL" or "FAILED"?

trippersham: 🦅👀

ecaoile (approved):

PR #90, opened 2021-02-24
Replace node-sass with sass

jchi2241:
## Summary
Based against https://github.com/HF-RapidResponse/volunteer-portal/pull/89

- `node-sass` is [deprecated](https://www.npmjs.com/package/node-sass) and replaced with `sass`. also speeds up client dependency installs by a lot
- `serve` changes:
  - move `serve` to devDependencies
  - change port so it doesn't conflict with create-react-app's default 8080
  - fix bug: https://stackoverflow.com/a/60514469 with flag `--no-clipboard`

## Test Plan
- npm run build, npm run serve, stuff still works
- make up, stuff still works

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -27,6 +27,7 @@
     "twix": "^1.3.0"
   },
   "scripts": {
+    "start": "npm run serve",
     "serve": "serve -s -n -l 5000 build",
     "client": "react-scripts start",
     "build": "react-scripts build",
--- a/client/package.json
+++ b/client/package.json
@@ -14,7 +14,6 @@
     "bootstrap": "^4.5.3",
     "lodash": "^4.17.20",
     "moment": "^2.29.1",
-    "node-sass": "^4.14.1",
     "react": "^16.14.0",
     "react-bootstrap": "^1.4.0",
     "react-dom": "^16.14.0",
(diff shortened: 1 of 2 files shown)


client/package.json
```diff
@@ -21,11 +20,10 @@
     "react-router-bootstrap": "^0.25.0",
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.0",
-    "react-share": "^4.3.1",
-    "serve": "^11.3.2"
+    "react-share": "^4.3.1"
   },
   "scripts": {
-    "start": "serve -s build",
+    "serve": "serve -s -n -l 5000 build",
```
jtuckerk, line 31: The app engine config that we use depends on there being a ["start" option](https://cloud.google.com/appengine/docs/standard/nodejs/runtime#application_startup) 
I do like this change though since it's more descriptive. 

We are moving away from app engine shortly so it's not a big deal, for now though could we include a duplicate start command?

jtuckerk (approved):
Thanks for making this change!

I've built this fresh and ran/tested it and verified it works

trippersham (approved):

client/package.json
jchi2241: yup for sure

jchi2241: should the port be defaulted back to 8080? not sure what the PORT env is to in the app engine

ecaoile (approved):

PR #91, opened 2021-02-24
Docs port warning

jtuckerk:
Open to other suggestions about how to make this more clear.

I think used the wrong port because I ran the "docker-compose up" command and then saw the client port info in my terminal and proceeded to test things from that port. So adding the appropriate port before the command seemed like one way to avoid using the wrong port.

We could also turn off the client port forwarding from docker.


jchi2241 (approved):
nice :) 

a couple other options to consider:
add prefixers to API paths in the client code that change depending on the NODE_ENV (altho i do like the cleanliness of just hitting `/api/...` rather than `${API_HOST}/api/...`)

we can also log into the terminal after client dev is up to visit localhost:8000 in an obvious way.

jtuckerk:
@vai0 All good suggestions! I keep starting on digging into them but get distracted with other more pressing things. going to leave this as is for now

PR #92, opened 2021-02-25
Support absolute imports

jchi2241:
## Summary
Support absolute imports. So we can do stuff like
```
import LoadingSpinner from "components/LoadingSpinner";
```
rather than
```
import LoadingSpinner from "../../components/LoadingSpinner";
```

see: https://create-react-app.dev/docs/importing-a-component/

## Test Plan
Made a change to an existing relative import, works

### DIFF
--- a/client/jsconfig.json
+++ b/client/jsconfig.json
@@ -1,0 +1,6 @@
+{
+  "compilerOptions": {
+    "baseUrl": "src"
+  },
+  "include": ["src"]
+}


jtuckerk (approved):
Thanks for including all that info in the comment! I am not a node expert so feel free to wait for Earl Jay's approval if youre not sure on this, but from what I read it seems good

ecaoile:
> Thanks for including all that info in the comment! I am not a node expert so feel free to wait for Earl Jay's approval if youre not sure on this, but from what I read it seems good

Doesn’t quite require node expertise lol. It’s just making imports a little easier

ecaoile (approved):

PR #93, opened 2021-02-25
Add Storybook support

jchi2241:
## Summary
This is based against https://github.com/HF-RapidResponse/volunteer-portal/pull/90

Add storybook support so we can begin documenting our components. 
This will help:
- better sync the design/eng team - e.g., what's already been built, have design get a better feel for low/high effort tasks.
- onboard new volunteers!

I'll eventually remove the examples storybook dumped into /stories in the next diff. They're there now so we know storybook works.

## Test Plan
1. `make shell-client`
2. `npm run storybook`


ecaoile (approved):
I've never used Storybook, but it looks legit.

PR #95, opened 2021-02-26
Implement account creation with CRUD functionality (oauth and regular)

ecaoile:
This is pretty much the entire accounts feature. Users can create accounts and manage their settings.


client/src/components/Header/index.js
```diff
@@ -0,0 +1,171 @@
...
+    history,
+    initiatives,
+    getInitiatives,
+    syncInitMapAndLoadUser,
+  } = props;
+
+  const firstPath = window.location.pathname;
+  const params = new URLSearchParams(window.location.search);
+
+  const mainLinks = [
```
jtuckerk, line 48: The new nav configuration is a bit squeezed on the left and theres a big gap on the right in a 1920x1080 screen
image

client/src/components/Register.js
```diff
@@ -1,122 +1,198 @@
...
+              placeholder="Password"
+              onChange={(e) => {
+                handleKeyPress('password', e);
+              }}
+              isInvalid={errors.password}
+              required
+            />
+            <Form.Control.Feedback type="invalid">
+              {
+                'Please enter a password between 6 and 20 characters long with at least 1 letter, 1 nuumber, and 1 special character.'
```
jtuckerk: 1 nuumber

jtuckerk: I'm getting this error with pass.1234 which I think should be valid

*time passes*

turns out I hadn't set up the db with the new tables. Could we separate this error from backend 500s in case we do experience 500s we dont want people thinking it's their faults

*more time passes* 

I believe I've got the backend setup properly and am not seeing any backend errors, but am still getting errors for  pass.1234 - also the check marks tell me its all good and the error message contradicts that. I think you brought that up in the meeting as an issue you were aware of.

api/models/account.py
```diff
@@ -0,0 +1,32 @@
...
+    first_name = Column('first_name', String(255), nullable=True)
+    last_name = Column('last_name', String(255), nullable=True)
+    password = Column('password', Text, nullable=True)
+    oauth = Column('oauth', String(32), nullable=True)
+    profile_pic = Column('profile_pic', Text, nullable=True)
+    city = Column('city', String(32), nullable=True)
+    state = Column('state', String(32), nullable=True)
+    roles = Column('roles', ARRAY(String), default=[], nullable=False)
+
+    UniqueConstraint(email, username, name='unique_acct_emails_usernames')
```
jtuckerk: This is redundant right? since the email and username are already independently unique? or does this somehow enforce that usernames can't be emails that already exist as emails?

api/models/account.py
```diff
@@ -0,0 +1,32 @@
...
+from sqlalchemy import func, select, Column, String, Integer, Text, Boolean, UniqueConstraint
+from sqlalchemy.dialects.postgresql import ARRAY, JSON, UUID
+from sqlalchemy.orm import backref, column_property, relationship, synonym
+from sqlalchemy.ext.hybrid import hybrid_property
+from pydantic import validator
+from sqlalchemy.sql import func
+from uuid import uuid4
+
+
+class Account(Base):
```
jtuckerk, line 12: before merging can you run `make db-reset-all-from-python` (and then add a new commit) this will update `db/data/dev/data.development.sql` and `db/data/test/data.test.sql` so the development DB is set up with all the proper tables when it is rebuilt

client/src/components/Header/MainMenu.js
```diff
@@ -0,0 +1,80 @@
...
+            }`}
+            key={`nav-dropdown-child-${j}`}
+            onClick={collapse}
+          >
+            {child.name || child.initiative_name}
+          </NavLink>
+        );
+      }
+      dropdownItems.push(
+        <NavDropdown.Divider
```
jtuckerk, line 29: I like this dropdown it looks nice and streamlines navigation, but I'm worried it will be really squeezed with all of our fairly long initiative names.

client/src/components/Register.js
```diff
@@ -1,122 +1,198 @@
...
-        </p>
+          </div>
+        </Form>
+        <p className="font-weight-bold side-line-text">or</p>
         <div className="text-center">
-          <Button onClick={SendDonorRequestLink}
-            variant="outline-info" style={{ padding: '.35rem 1.5rem' }}>
-            Submit Request
-          </Button>
+          <GoogleOAuthButton />
```
jtuckerk: Can this be at the top? I didnt see it and started to fill out my info even though I would have just preferred to use google in the first place if I had seen it.

client/src/components/Account/AccountNav.js
```diff
@@ -0,0 +1,95 @@
+import React, { useState } from 'react';
```
jtuckerk: The account/profile flows are all so great!

client/src/components/Account/Profile.js
```diff
@@ -0,0 +1,268 @@
...
+                isInvalid={errors.email}
+                readOnly
+              />
+              <Form.Control.Feedback type="invalid">
+                Email is invalid!
+              </Form.Control.Feedback>
+            </Form.Group>
+          </Col>
+          <Col xs={12} md={4} className="align-self-center pl-lg-5">
+            <Form.Switch
```
jtuckerk: Can we make email default to not showing on the profile? I doubt people want to display their emails by default, and they may not notice it is checked to display

api/models/account.py
```diff
@@ -0,0 +1,32 @@
...
+    __tablename__ = 'accounts'
+
+    uuid = Column(UUID(as_uuid=True), primary_key=True,
+                  default=uuid4, unique=True, nullable=False)
+    email = Column('email', Text, unique=True, nullable=False)
+    username = Column('username', String(255), unique=True, nullable=False)
+    first_name = Column('first_name', String(255), nullable=True)
+    last_name = Column('last_name', String(255), nullable=True)
+    password = Column('password', Text, nullable=True)
```
jtuckerk, line 23: Can we add a comment here noting that passwords are encrypted and a pointer to that logic?

api/models/settings.py
```diff
@@ -0,0 +1,29 @@
...
+from sqlalchemy import func, select, Column, String, Integer, Text, Boolean
+from sqlalchemy.dialects.postgresql import ARRAY, JSON, UUID
+from sqlalchemy.orm import backref, column_property, relationship, synonym
+from sqlalchemy.ext.hybrid import hybrid_property
+from pydantic import validator
+from sqlalchemy.sql import func
+from uuid import uuid4
+
+
+class Settings(Base):
```
jtuckerk: Could we maybe call this AccountSettings? just in case we have some other types of settings? I've been tossing around some ideas for an Admin api that might need it's own settings. we can/should call those adminSettings, but it might still save some confusion

jtuckerk (changes requested):
This is looking super good! I did a mostly superficial review so far and will take another pass in the next day or 2 to go through the code more closely

api/api.py
```diff
@@ -25,19 +29,24 @@
...
 
-app.add_middleware(SessionMiddleware, secret_key=Config['auth']['jwt']['secret'])
+app.add_middleware(SessionMiddleware,
+                   secret_key=Config['auth']['jwt']['secret'])
 app.include_router(
     auth.router,
     prefix='/api'
 )
 
+CLIENT_ID = "899853639312-rluooarpraulr242vuvfqejefmg1ii8d.apps.googleusercontent.com"
```
jtuckerk: can this be removed?

api/models/account.py
```diff
@@ -0,0 +1,32 @@
...
+    oauth = Column('oauth', String(32), nullable=True)
+    profile_pic = Column('profile_pic', Text, nullable=True)
+    city = Column('city', String(32), nullable=True)
+    state = Column('state', String(32), nullable=True)
+    roles = Column('roles', ARRAY(String), default=[], nullable=False)
```
jtuckerk, line 28: Is the plan to just have these be strings users can add themselves? or do we want roles to be tied to more functionality

api/models/account.py
ecaoile: They should be unique, but the constraint doesn't seem to work. I added constraints here and in the SQL file, but neither of them did nothing. Maybe I'm misunderstanding how UniqueConstraint is supposed to work. I ended up having to write in checks on the API to check for existing usernames and emails

client/src/components/Header/MainMenu.js
ecaoile: Do you want to get rid of it then?

client/src/components/Register.js
ecaoile: I suppose we could, but I think I'm used to seeing it at the bottom of forms. Have you seen it anywhere at the top? I suppose it might make sense to have it at the time for register at least, but I dunno...hmm...

client/src/components/Header/index.js
ecaoile: We can center it with ml-auto. How does this look?
image

ecaoile: Actually, I think the intent from the UI/UX folks was there to be a gap at least at bigger resolutions. We could probably add a little more margin-left, though.

client/src/components/Register.js
ecaoile: Ok, it was a combination of things. The regex for password validation was a little off: it didn't include periods or a few other characters. Oops. Also figured out how to disable the green checkmark when I don't need it. Wish Bootstrap didn't show the valid when an invalid prompt was present. Anyway, fix incoming...

ecaoile: Nm, I kinda lied. Password validation is busted and always shows valid feedback with any character. I dunno why.

client/src/components/Header/index.js
ecaoile: added a tiny bit of spacing:
image

client/src/components/Register.js
ecaoile: fixed

client/src/components/Account/AccountNav.js
ecaoile: Thanks, I aim to please!

api/api.py
ecaoile: Yes. Fixed.

api/models/settings.py
ecaoile: changed to AccountSettings

client/src/components/Account/Profile.js
ecaoile: fixed

api/models/account.py
ecaoile: That command doesn't work.
`
make: *** No rule to make target 'db-reset-all-from-python'.  Stop.`

No wait, I lied. I did this and now the file is empty. I don't think this is what you wanted.

api/config.yaml
```diff
@@ -6,8 +6,12 @@ production_db_defaults: &production_db_defaults
...
     host:
     database:
-    user:
-    password:
+      adapter: postgresql
+      generator: secret_generator
+      host: 10.17.1.3
+      database: hf_volunteer_portal
+      user: admin
+      password: humanity-forward_hf-portal_psql-db_admin
```
ecaoile: @jtuckerk This build is now busted after adding these config.yaml settings

api/settings.py
```diff
@@ -62,8 +62,11 @@ def read_config():
 # Generate Database URLs based on the DB connection information
 
 
-def generate_hf_mysql_db_address(connection) -> str:
-    db_pass = get_secret_for_key(db_secret_key)
+def generate_hf_secret_db_address(connection) -> str:
```
ecaoile, line 65: @jtuckerk this might have broken it too

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
+
 @app.post("/api/donation_link_requests/", response_model=DonationEmailSchema)
 def create_donation_link_request(donationEmail: DonationEmailSchema, db: Session = Depends(get_db)) -> DonationEmailSchema:
     db.add(DonationEmail(email=donationEmail.email))
     db.commit()
     return donationEmail
+
+
+@app.get("/api/accounts/", response_model=List[AccountResponseSchema])
+def get_all_accounts(db: Session = Depends(get_db)):
```
jtuckerk: Dont forget to delete this! We need an admin API to authorize stuff like this

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
+
+@app.get("/api/accounts/email/{email}", response_model=AccountResponseSchema)
+def get_account_by_email(email, db: Session = Depends(get_db)):
+    return db.query(Account).filter_by(email=email).first()
+
+
+@app.post("/api/accounts/", response_model=AccountResponseSchema, status_code=201)
+def create_account(account: AccountRequestSchema, Authorize: AuthJWT = Depends(), db: Session = Depends(get_db)):
+    existing_acct = db.query(Account).filter_by(email=account.email).first()
+    if existing_acct is not None:
```
jtuckerk: I'm looking into the db unique constraint stuff and I think we can get the DB to check the uniqueness of emails itself. tho this might be an easier way than trying to handle the error from the db.

I've sent PR#104 with tests and things related to this

api/schemas/account.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import List, Dict, Text, Optional
+from uuid import UUID
+
+
+class AccountRequestSchema(BaseModel):
+    email: EmailStr
+    username: Optional[str]
+    first_name: str
+    last_name: str
+    password: Optional[Text]
```
jtuckerk, line 11: I think we'll want to exclude this from the schema

jtuckerk (commented):
I really dont like how github structures these reviews. It seems like I have to submit this before responding to your responses? more comments coming in a sec

api/models/account.py
jtuckerk: I've added some other comments about this and PR#104. I still dont think the check in api.py is a bad idea, but we will need to do that check for username also so maybe it is easier to just let the db do that check not that (i think) we have a fix for that

client/src/components/Register.js
jtuckerk: ah I thought I tried an @ symbol. maybe that was one of the other non-special special characters

api/models/account.py
jtuckerk: hmmm that's odd. Did you do `docker-compose down` first? I just did it in PR#104 and it seemed to work as expected

client/src/components/Register.js
jtuckerk: hmm I think youre right, it is usually at the bottom, but it's usually not off the edge so I can still see it, maybe we could have the self account creation fields expand if that's selected?

api/config.yaml
jtuckerk: ahh heck. I'll look into it

api/models/account.py
ecaoile: I honestly don't know what the intended functionality will be for roles, so I just guessed that we would be able to put them into an array. We might actually have to do something like the initiative_map that I moved to acct settings.

api/api.py
ecaoile: Oops, yeah, I made that route for testing. We really shouldn't be leaving that open for everyone.

api/schemas/account.py
ecaoile: Wait, if we exclude properties from the schema, are we still able to pass them in to the API?

client/src/components/Register.js
ecaoile: We could have it similar to what I did for the password change where you don't show the new password and retype until someone actually types in. To make it more obvious there's more info to put in, we could add some kind of opacity fade out effect with CSS at the bottom as a preview.

jtuckerk:
I'm getting "welcome back, Undefined!" when I started up the app. I'm guessing that has something to do with having a token in my browser but the db being blank? This probably wont come up in prod, unless someone forced it, but I'm wondering if its something we should bother preventing
image

client/src/components/Header/index.js
ecaoile: Spacing has been pushed, so I'll resolve this too unless you want more.

client/src/components/Register.js
ecaoile: Yeah, I suck at regex. Anyway, . and @ are both working now.

ecaoile:
> 
> 
> I'm getting "welcome back, Undefined!" when I started up the app. I'm guessing that has something to do with having a token in my browser but the db being blank? This probably wont come up in prod, unless someone forced it, but I'm wondering if its something we should bother preventing
> image

Oh, we definitely should be checking for that, though. That shouldn't be too hard to fix. We can just include cookies into the parameter, and if the user doesn't exist in the db, we just remove that cookie and refresh the page.

Update: this has been addressed.

client/src/components/Register.js
ecaoile: I fixed the password issue. It was really was regex. Bob Saget, Kelly Clarkson, Ryan Seacrest!

client/src/components/OldRegister.js
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
```
jtuckerk: Should we just delete this file?

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
     db.commit()
     return donationEmail
+
+
+@app.get("/api/accounts/", response_model=List[AccountResponseSchema])
+def get_all_accounts(db: Session = Depends(get_db)):
+    return db.query(Account).all()
+
+
+@app.get("/api/accounts/{uuid}", response_model=AccountResponseSchema)
```
jtuckerk: could we put al the account routes into their own account_api.py file? like the auth.py with auth.router

api/auth.py
```diff
@@ -84,18 +169,32 @@ async def authorize_github(request: Request, Authorize: AuthJWT = Depends()):
...
-    # refresh_token = Authorize.create_refresh_token(subject=user_id)
-    response = RedirectResponse(f'{Config["routes"]["client"]}/auth_callback')
+    refresh_token = Authorize.create_refresh_token(subject=user_id)
     Authorize.set_access_cookies(access_token, response)
-    # Authorize.set_refresh_cookies(refresh_token)
+    Authorize.set_refresh_cookies(refresh_token, response)
     return response
 
+
+def create_access_and_refresh_tokens(user_id: str, Authorize: AuthJWT):
```
jtuckerk: This is the second instance  of this function, can we just import it from here into api.py?

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
+def get_account_by_uuid(uuid, Authorize: AuthJWT = Depends(), db: Session = Depends(get_db)):
+    Authorize.jwt_required()
```
jtuckerk: This just ensures that they have a valid jwt, but doesnt actually prevent people from getting other peoples' account info.

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
+    return db.query(Account).filter_by(uuid=uuid).first()
+
+
+@app.get("/api/accounts/email/{email}", response_model=AccountResponseSchema)
+def get_account_by_email(email, db: Session = Depends(get_db)):
+    return db.query(Account).filter_by(email=email).first()
```
jtuckerk: Auth! fine to leave all the auth stuff for a later PR, but can we add TODOs so we dont forget/people know its intentionally left out for now?

api/auth.py
```diff
@@ -65,14 +76,88 @@ async def login(request: Request, provider: OAuthProvider):
...
+    password_valid = check_encrypted_password(
+        payload.old_password, existing_acct.password)
+    if password_valid is True:
+        return True
+    else:
+        raise HTTPException(
+            status_code=403, detail=f"Password is is incorrect")
+
+
+@router.post("/auth/basic")
```
jtuckerk, line 126: this wil be the route /auth/auth/basic -- there is an auth/auth/google above so maybe that's the way it's supposed to be?

api/bootstrap.py
```diff
@@ -26,7 +26,8 @@
 # Create database(s) and tables
 for key in Connections:
     # Connect with `echo` so we can see what's being run
-    db_url = Connections[key]['url'].replace(f'/{Connections[key]["database"]}', '/postgres')
+    db_url = Connections[key]['url'].replace(
```
jtuckerk: this replace call can just be removed. it was forcing all dbs to postrgress, but I dont think it really even makes sense any more given the shape of the configs

api/config.yaml
```diff
@@ -1,82 +1,86 @@
 # Aliases for common sections of configuration across environments
 
 production_db_defaults: &production_db_defaults
+  database:
+    adapter: mysql
+    generator: secret_generator
+    host:
```
jtuckerk: I dont think we need these? and if we do something else is wonky.

api/schemas/account.py
```diff
@@ -0,0 +1,44 @@
...
+    uuid: UUID
+
+
+class AccountBasicLoginSchema(BaseModel):
+    email: EmailStr
+    password: Text
+
+
+class AccountPasswordSchema(BaseModel):
+    old_password: Text
```
jtuckerk, line 33: does this need to be "old"?

api/schemas/account.py
```diff
@@ -0,0 +1,44 @@
...
+    uuid: UUID
+
+
+class PartialAccountSchema(BaseModel):
```
jtuckerk, line 37: was above class going to be used for the /api/accounts/{uuid} updated_password route? seems like for now that class having uuid and password should be sufficient.

api/api.py
```diff
@@ -51,32 +58,171 @@ def authjwt_exception_handler(request: Request, exc: AuthJWTException):
...
+
+def check_for_diff_user_with_same_username(uuid, account: AccountRequestSchema, db: Session):
+    existing_acct = db.query(Account).filter_by(
+        username=account.username).first()
+    if existing_acct is not None and str(uuid) != str(existing_acct.uuid):
+        raise HTTPException(status_code=400,
+                            detail=f"Account with username {existing_acct.username} already exists!")
+
+
+@app.patch("/api/accounts/{uuid}", response_model=AccountResponseSchema)
```
jtuckerk: Is patch typically used for just password changes? would someone well versed in rest/http things know from this url and the patch that this only updates the password for an account?

api/models/account_settings.py
```diff
@@ -0,0 +1,29 @@
...
+                  default=uuid4, unique=True, nullable=False)
+    show_name = Column('show_name', Boolean, default=False, nullable=False)
+    show_email = Column('show_email', Boolean, default=False, nullable=False)
+    show_location = Column('show_location', Boolean,
+                           default=True, nullable=False)
+    organizers_can_see = Column(
+        'organizers_can_see', Boolean, default=True, nullable=False)
+    volunteers_can_see = Column(
+        'volunteers_can_see', Boolean, default=True, nullable=False)
+    initiative_map = Column('initiative_map', JSON, default={}, nullable=False)
```
jtuckerk, line 26: could you add a comment indicating what the keys and values of this map are? initiative_name -> isSubscribed?

I think we'll want to talk with @trippersham about how hes setting up the subscriptions since I think he's got a different plan in mind. From what I understand the subscriptions will be their own DB where it's an account foreign key is included and each row represents a subscription so that the same structure can be used for accounts and non-account based subscriptions

api/security.py
```diff
@@ -0,0 +1,15 @@
...
+
+pwd_context = CryptContext(
+    schemes=["pbkdf2_sha256"],
+    default="pbkdf2_sha256",
+    pbkdf2_sha256__default_rounds=30000
+)
+
+
+def encrypt_password(password):
+    return pwd_context.encrypt(password)
```
jtuckerk: Getting a deprecation warning for this in the tests:

/api/security.py:11: DeprecationWarning: the method passlib.context.CryptContext.encrypt() is deprecated as of Passlib 1.7, and will be removed in Passlib 2.0, use CryptContext.hash() instead.
    return pwd_context.encrypt(password)

client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "volunteer-portal",
-  "version": "0.5.0",
+  "version": "1.0.0",
```
jtuckerk, line 3: nice

client/package-lock.json
```diff
@@ -10666,11 +10883,11 @@
       }
     },
     "file-entry-cache": {
-      "version": "6.0.1",
-      "resolved": "https://registry.npmjs.org/file-entry-cache/-/file-entry-cache-6.0.1.tgz",
-      "integrity": "sha512-7Gps/XWymbLk2QLYK4NzpMOrYjMhdIxXuIvy2QBsLE6ljuodKvdkWs/cpyJJ3CVIVpH0Oi1Hvg1ovbMzLdFBBg==",
+      "version": "5.0.1",
```
jtuckerk: I'm seeing a few packages that look downgraded, think that's on purpose or a merge thing?

api/models/account.py
```diff
@@ -0,0 +1,31 @@
...
+    email = Column('email', Text, unique=True, nullable=False, index=True)
+    username = Column('username', String(255), unique=True, nullable=False, index=True)
+    first_name = Column('first_name', String(255), nullable=True)
+    last_name = Column('last_name', String(255), nullable=True)
+    # passwords are encrypted using passlib with pbkdf2_sha256 hashing (see security.py)
+    password = Column('password', Text, nullable=True)
+    oauth = Column('oauth', String(32), nullable=True)
+    profile_pic = Column('profile_pic', Text, nullable=True)
+    city = Column('city', String(32), nullable=True)
+    state = Column('state', String(32), nullable=True)
```
jtuckerk, line 27: Can we  start off including zipcode as well (or instead?) we've had some volunteers do a bunch of work to get zip code to federal and state congressional districts mappings and its also pretty easy to map zip to city and state if we want

client/src/components/Account/AccountNav.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const leftArrowClick = () => {
+    const linksCopy = [...mobileLinks];
+    linksCopy.push(linksCopy.shift());
+    setMobileLinks(linksCopy);
+  };
+
+  return (
+    <>
+      <Nav
```
jtuckerk: I've spun up a staging version of this PR [here](https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app/) so that the UX folks can take a look. I'd like to get their input on his nav setup and a couple of other things.

client/src/components/Header/LoggedInMenu.js
```diff
@@ -0,0 +1,61 @@
...
+      <NavDropdown
+        title={`Welcome back, ${user.first_name}!`}
+        key={`nav-top-profile`}
+      >
+        {profileDropdown}
+      </NavDropdown>
+      <Image
+        src={
+          user.profile_pic ||
+          (user.username === 'andyfromtheblock' ? andyPic : placeholderImg)
```
jtuckerk, line 52: This probably doesnt matter, but did you intend to leave this in? we could hard code these values into the DB config?

client/src/components/Register.js
ecaoile: moved oauth to top

api/api.py
ecaoile: commented out

api/models/account.py
ecaoile: Resolving this for now. We can change the prop later if needed

api/models/account.py
ecaoile: It works now.

powell759:
Taking a first functional pass - looks fantastic!

Size is a bit fatiguing though; since this is security-critical functionality I'm going take several passes. Would recommended keeping PRs much smaller going forward.

powell759:
image

Is it possible to have human-readable name here?

powell759:
functional nit:

Password requirements are not enforced by API, run following from console:

```
fetch("http://localhost:8000/api/accounts/", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-US,en;q=0.9",
    "content-type": "application/json;charset=UTF-8",
    "sec-ch-ua": "\"Google Chrome\";v=\"89\", \"Chromium\";v=\"89\", \";Not A Brand\";v=\"99\"",
    "sec-ch-ua-mobile": "?0",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-origin",
    "sec-gpc": "1"
  },
  "referrer": "http://localhost:8000/register",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": "{\"username\":\"foo\",\"email\":\"bar@fuzz.com\",\"first_name\":\"test\",\"last_name\":\"tester\",\"password\":\"password\",\"roles\":[]}",
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
```

powell759:
Question on API - I'm seeing a "roles" field for account creation requests. If this is for admin/privileges? If so then we'll need restrictions here to prevent privilege escalation.

Example of arbitrary role creation:
```
fetch("http://localhost:8000/api/accounts/", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-US,en;q=0.9",
    "content-type": "application/json;charset=UTF-8",
    "sec-ch-ua": "\"Google Chrome\";v=\"89\", \"Chromium\";v=\"89\", \";Not A Brand\";v=\"99\"",
    "sec-ch-ua-mobile": "?0",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-origin",
    "sec-gpc": "1"
  },
  "referrer": "http://localhost:8000/register",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": "{\"username\":\"test1\",\"email\":\"test1@test.com\",\"first_name\":\"test\",\"last_name\":\"tester\",\"password\":\"password\",\"roles\":[\"Test Role\"]}",
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
```

ecaoile:
> 
> 
> image
> 
> Is it possible to have human-readable name here?

I think this is sometimes that needs to be changed on the Google end. @trippersham can you change the name?

ecaoile:
> 
> 
> Question on API - I'm seeing a "roles" field for account creation requests. If this is for admin/privileges? If so then we'll need restrictions here to prevent privilege escalation.
> 
> Example of arbitrary role creation:
> 
> ```
> fetch("http://localhost:8000/api/accounts/", {
>   "headers": {
>     "accept": "application/json, text/plain, */*",
>     "accept-language": "en-US,en;q=0.9",
>     "content-type": "application/json;charset=UTF-8",
>     "sec-ch-ua": "\"Google Chrome\";v=\"89\", \"Chromium\";v=\"89\", \";Not A Brand\";v=\"99\"",
>     "sec-ch-ua-mobile": "?0",
>     "sec-fetch-dest": "empty",
>     "sec-fetch-mode": "cors",
>     "sec-fetch-site": "same-origin",
>     "sec-gpc": "1"
>   },
>   "referrer": "http://localhost:8000/register",
>   "referrerPolicy": "strict-origin-when-cross-origin",
>   "body": "{\"username\":\"test1\",\"email\":\"test1@test.com\",\"first_name\":\"test\",\"last_name\":\"tester\",\"password\":\"password\",\"roles\":[\"Test Role\"]}",
>   "method": "POST",
>   "mode": "cors",
>   "credentials": "include"
> });
> ```

I'm not sure what the intended purpose of roles is yet. It was just something included in the mock-up. I initialized it as an empty array so that we could potentially add dummy strings to it for mocking.

powell759:
We're exposing private information to anyone who knows an email address through the "Get Account By Email" API. Is there a reason that this is public facing?

ecaoile:
> 
> 
> We're exposing private information to anyone who knows an email address through the "Get Account By Email" API. Is there a reason that this is public facing?

I'll double check, but I think it's not being used anymore. E-mail check is being done in the other endpoints now. If that's the case, I'll delete it.

ecaoile:
Password format checks have been added to post and patch. Get account by e-mail endpoint deleted.

powell759:
> > We're exposing private information to anyone who knows an email address through the "Get Account By Email" API. Is there a reason that this is public facing?
> 
> I'll double check, but I think it's not being used anymore. E-mail check is being done in the other endpoints now. If that's the case, I'll delete it.

Sounds good - it was also exposing UUID allowing me to delete / update other accounts

jtuckerk (changes requested):
I've finally gone through this line by line! you did so much!

sorry for the delay on this, I was trying to finish up yesterday but didnt quite get the deployed version working. Its up now at https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app/ there is  an issue with the routes so the google login isnt working, ~I think it has to do with the change from 8080 in the config to 8000, but I'm verifying that now. The manual account creation works for now tho.~ @trippersham I think you had asked if the staging url is constant, It is, and I guess we need to include it somewhere in the config for this to send the token back to the backend

Theres some overlap in mine and max's comments but I've just left them. I think we dont need to make all the changes to the api just yet to finish up fully securing all of the calls. We can gets trips input since he set up the auth and this will just become too big of a PR.

I just have a few things I want the UX peeps to look at which is why ive put it up on that link above, and a handfull of minor change requests

api/config.yaml
jtuckerk: checked in branch release-staging-ux-review, this change breaks the configs. that branch has a working config.

api/api.py
ecaoile: I got rid of it.

client/src/components/Header/LoggedInMenu.js
ecaoile: Oh, that was just a funny easter egg since I wanted to test accounts with multiple pictures. I don't think many people are gonna want to go by andyfromtheblock. Are we going to integrate picture upload in the future?

api/api.py
powell759: +1 and in general let's keep UUID (and database keys in general) off the client side / out of responses. Can we associate tokens with accounts on the backend and use them (rather than UUID) for state-changing operations?

api/api.py
ecaoile: I was scared it was going to break if I tried to make it modular, but yes, we should do that. Api.py is getting too fat.

client/package-lock.json
ecaoile: Merge blues.

client/src/components/OldRegister.js
ecaoile: I don't know if we were gonna use any of this later. Figure we might at least have use for that social share.

api/api.py
ecaoile: I don't know if that's legal, but I'll see what I can do.

client/src/components/Header/LoggedInMenu.js
jtuckerk: Im fine with easter eggs
yeah I think we should, makes it feel more like a community

jtuckerk:
re: the bit about auth not working in staging. that's more complicated and we can deal with that outside of this PR. 

I think we'll want to include `proxy_set_header Host $host;` for the api in the nginx conf since I'm seeing that the auth redirect is trying to hit the docker internal http:/api:8081 domain, but just adding that messes up other things so theres more digging that needs to be done. I think @trippersham  has already grappled with this problem a bit

api/api.py
ecaoile: Ok, figured it out

api/api.py
ecaoile: Looks like it might be legal as long as the uuid is stored in the token somewhere. Don't think we can completely hide it from the front end, but we can make it less visible. On the other hand, I don't know how we're going to use cookies now if I can't hold the UUID in them

powell759: @jtuckerk's point is that while we are validating the session JWT we aren't validating that it is that user's data they are accessing. This means that any user with a valid JWT can preform CRUD operations on any user's account. What we need to prevent that is to have an association between session and account, and we need to validate that on requests (At some point we may want privileged accounts but we can cross that bridge when we come to it).

My point was that UUID shouldn't be on the client side or used in the path for APIs at all. Exposing database keys increases blast radius of any potential attack. In Tucker's example a user can update another person's information - if UUID wasn't an argument that wouldn't be possible.

We don't need to worry about storing UUID in the token and I wouldn't recommend that anyway, I believe that this JWT is only signed - not encrypted - and the info could be extracted. Take a look at the "Get All Settings" API `/api/account_settings/` - this takes no argument yet returns the user's account. That tells me we already are making an association on the backend between the JWT session and a user.

All access/updates should function like this endpoint so we have:
- a JWT required to access
-  resource accessed defined by session rather than UUID

ecaoile: I get what Tucker was saying, but I'm just unsure how to get the proper user credentials if we don't pass some kind of primary key even if it isn't the UUID. It's simple enough to get the jwt subject and ensure that it's equivalent to the current user, but if we don't use UUID, what do we use? We need a unique identifier, so it would have to be e-mail or username. In that case, wouldn't that key serve the same purpose as UUID even if we don't pass it into the API endpoint address? 

The get all for account_settings actually returns settings from all users. This needs to be disabled too until we configure admin privileges.

api/auth.py
```diff
@@ -65,37 +76,144 @@ async def login(request: Request, provider: OAuthProvider):
...
+
+    if account is None:
+        new_account = Account(
+            email=email_to_use,
+            username=user['login'],
+            first_name=user['name'],
+            last_name='no last name',
+            oauth='github',
+            profile_pic=user['avatar_url'],
+            city=user['location'].split(', ')[0],
```
jtuckerk: I dont have my location set for github so this fails:
```
api_1     |     return await dependant.call(**values)
api_1     |   File "./auth.py", line 182, in authorize_github
api_1     |     city=user['location'].split(', ')[0],
api_1     | AttributeError: 'NoneType' object has no attribute 'split'
```

ecaoile: Oof, I see GitHub makes a lot of the fields optional. Probably best to always do a null check for string splits anyway. Silly me.

api/models/account.py
ecaoile: added zip code to front end and back end

api/api.py
powell759: Okay - took a closer look this evening - I'm familiar with session / cookie based authentication where you'd provide a user with an expiring session and associate that to the user on the backend. That way you'd never have the need for a user id on the client.

Did some learning and it looks like the JWT / token-based authentication is a whole different approach with one of the main benefits being not needing that datastore for sessions. So implementing a lookup kinda gets rid of that benefit. So if we are going with a token-based authentication then you are right, we'll need to include some type of unique key, and whether or not that is UUID doesn't really matter (it looks UUID is already the subject claim so might as well leave it).

I think there are some minor drawbacks to using JWT like not being able to revoke authorizations - but I'm not sure we'll really have a use case for that.

So nvm the UUID comment - let's just add the subject check and disable the all accounts API.

ecaoile: Added token check on all the account endpoints except the posts. I feel like there should be some kind of check for post too, but I don't know what the best practice is when we haven't created a user yet. I wonder if the app should issue some kind of non-account jwt so that it knows the request was being made from the app.

api/auth.py
ecaoile: It's using api as a prefix, not auth.You can see on the front end that it's calling /api/auth/basic

api/schemas/account.py
ecaoile: I think originally I was planning to use this class for patch in general (i.e. making a change to one prop), but something about it was wonky. I forgot off the top of my head. Maybe it had something to do with not knowing how to get the props that were changed...or maybe it was the password. If you know how to properly do a patch with SQL Alchemy, maybe you could help me out here.

api/schemas/account.py
ecaoile: I guess we could call it current_password, but we're trying to get rid of it. Is old_password misleading?

api/auth.py
ecaoile: deleted and importing from account_api

api/config.yaml
ecaoile: Grabbed the config from that branch.

api/security.py
ecaoile: changed to .hash()

api/bootstrap.py
ecaoile: So just leave it as `db_url = Connections[key]['url']`?

api/models/account_settings.py
ecaoile: Yes. A lot of this data structure is guessing since I don't know for sure what we're doing with some of these props. I think I know what you mean with that subscription and account table, though We use a few of those at my workplace. I guess for that table, we could use e-mails so that it works for account and non-account subs.

api/auth.py
```diff
@@ -179,8 +179,10 @@ async def authorize_github(request: Request, Authorize: AuthJWT = Depends(), db:
             last_name='no last name',
             oauth='github',
             profile_pic=user['avatar_url'],
-            city=user['location'].split(', ')[0],
-            state=user['location'].split(', ')[1],
+            city=None if user['location'] is None else user['location'].split(', ')[
```
jtuckerk, line 183: could do `user.get('location', None)` instead of `user['location']` just in case the field does not exist. but if the api is currently sending none, I'd be surprised if they started dropping the field entirely, so its probably fine as is.

client/package-lock.json
jtuckerk: oh :( is there a simple way to update everything to the latest version?

ecaoile: I'm not sure if it's really an issue. It's possible I accidentally downgraded one of the packages in packages.json after merging previously, but we typically don't tamper with the package-lock file much. That just tends to get generated after adding the dependencies and then running npm install.

I ran npm uninstall and tried re-installing for good measure, but it didn't appear to change anything.

client/src/components/OldRegister.js
ecaoile: Nevermind, I see we're getting rid of that donation link thing. I'll just delete this then.

client/package-lock.json
jtuckerk: ok that's good enough for me!

client/src/components/Account/AccountNav.js
jtuckerk: Their comments can be addressed in later PRs, resolving

api/api.py
ecaoile: Going to resolve this too. Post endpoints are still kinda sus, but we can improve on them later.

api/auth.py
```diff
@@ -65,37 +77,134 @@ async def login(request: Request, provider: OAuthProvider):
...
     try:
         token = await oauth.github.authorize_access_token(request)
     except OAuthError as error:
         return HTMLResponse(f'<h1>{error.error}</h1>')
     resp = await oauth.github.get('user', token=token)
     user = resp.json()
-    return create_token_for_user(Authorize, user['login'])
-
-@router.delete("/logout")
+    email_to_use = user['email'] or user['login'] + '@fakegithubemail.com'
```
jtuckerk, line 171: Could we include the username in a message: "Github auth account: {username}" or something similar to avoid confusion?

api/account_api.py
```diff
@@ -0,0 +1,186 @@
...
+    acct_to_delete = db.query(Account).filter_by(uuid=uuid).first()
+    if acct_to_delete is None:
+        raise HTTPException(status_code=400,
+                            detail=f"Account with UUID {uuid} not found")
+    db.delete(acct_to_delete)
+    db.commit()
+
+
+@router.post("/account_settings/", response_model=AccountSettingsSchema, status_code=201)
+def create_settings(settings: AccountSettingsSchema, Authorize: AuthJWT = Depends(), db: Session = Depends(get_db)):
```
jtuckerk, line 139: Any way we can make this dependent on account creation? could we create the settings in the create account call and have an authed request later to get the settings or we do an authed request to this method after account creation

api/auth.py
jtuckerk: yeah I was confused. I thought the api prefix was we had was /api/auth for some reason

jtuckerk: Wondering if we should organize them that way. might make the auto generated docs a little cleaner. can be a later PR.

api/auth.py
```diff
@@ -65,37 +77,134 @@ async def login(request: Request, provider: OAuthProvider):
...
+                1],
+        )
+        db.add(new_account)
+        db.commit()
+        db.refresh(new_account)
+        account = new_account
+    return create_token_for_user(Authorize, str(account.uuid))
+
+
+@ router.delete("/logout")
```
jtuckerk: delete space? @ router (surprisingly doesnt seem to matter)

jtuckerk (approved):
Good to go! 
there are just a few things we'll probably want to do in followup CLs, I've aggregated them here:

- We'll need to force the email duplicate checker to compare in a case insensitive manner. right now I can create Tucker@gmail and tucker@gmail.
- also We should remove periods from before the @ sign before checking. [dots don't matter in gmail](https://support.google.com/mail/answer/7436150?hl=en). I assume some other email providers do the same.
- Right now for the login and create buttons, the oauth login flows both create an account if one does not exist. I imagine that would confuse people if they have an account but forget how they signed up. if they try logging in with gmail it will appear we've wiped out their settings.
- fakegithubemail - can this look different?
- account API auth finishing touches/tests
- separate various account routes with additional path names for a more hierarchical structure.
- route for the logout delete route -- just a lint thing

jtuckerk:
My review was very light on the react stuff, but I think we can merge without a react review unless @powell759, @vai0, or you feel differently

api/auth.py
ecaoile: image

Disabled username change for GitHub users and wrote a note for fake e-mails.

api/auth.py
ecaoile: space deleted

ecaoile:
> 
> 
> Good to go!
> there are just a few things we'll probably want to do in followup CLs, I've aggregated them here:
> 
>     * We'll need to force the email duplicate checker to compare in a case insensitive manner. right now I can create Tucker@gmail and tucker@gmail.
> 
>     * also We should remove periods from before the @ sign before checking. [dots don't matter in gmail](https://support.google.com/mail/answer/7436150?hl=en). I assume some other email providers do the same.
> 
>     * Right now for the login and create buttons, the oauth login flows both create an account if one does not exist. I imagine that would confuse people if they have an account but forget how they signed up. if they try logging in with gmail it will appear we've wiped out their settings.
> 
>     * fakegithubemail - can this look different?
> 
>     * account API auth finishing touches/tests
> 
>     * separate various account routes with additional path names for a more hierarchical structure.
> 
>     * route for the logout delete route -- just a lint thing

I've added form sanitization with space trim for some fields and lowercase + period removal for e-mails. Also hiding e-mails with message for GitHub users. I'll leave this PR up for a few hours in case anyone else wants to review, and then I'll merge it later in the evening.

api/account_api.py
ecaoile: Probably, but that will require adding some default dummy values into a class like I did on the JS side. If I could do that, I wouldn't have to change any other code besides adding the require JWT and making the call from the account post to the settings. I'll address this in the next PR.

PR #100, opened 2021-03-04
Commit airtable rows individually and propogate error info to api.

jtuckerk:
Since the DB now defines some columns as non-nullable, certain rows in the events, roles, or initiative tables are now considered malformed and cause the whole sync to fail. (this is not how it is set up in the current mysql table)

I've  split each row add into its own transaction, which I didnt want to do, but there was no way to force SQLAlchemy to validate that the non-null conditions were satisfied on an ORM object prior to committing, without adding my own separate validation wrapper around the existing objects. Since this isnt directly user facing and the tables are small, I dont think this will have a big impact (though I havent checked and will do that shortly).

This change also adds in some code to bring the individual errors to the api and presents them with a bit of HTML formatting to make debugging easier.


ecaoile:
This branch is a bit stale. Are we still trying to merge it?

jtuckerk:
> This branch is a bit stale. Are we still trying to merge it?

yes! I've just caught it back up with develop

ecaoile (approved):

PR #101, opened 2021-03-04
remove donation link request from client, api, and db.

jtuckerk:
it was a rushed implementation and I dont see a good reason for
  keeping it around. If we want something similar we can bring it back
  more intentionally. maybe include it into the accounts. also automate
  the response.


powell759 (approved):
LGTM  - ship it.

PR #104, opened 2021-03-10
Add tests for account creation

jtuckerk:
I've added some tests for account things I wanted to dig into. 
I think maybe the reason you didnt see the unique column constraint working might have had to do with using an old db schema from before you added unique=True. 

I've also made username and email indices so that the check for duplicates will be faster.

I verified by commenting out lines 119-122 in api.py (not included here) that the db integrity error is triggered on duplicate emails


api/models/account.py
```diff
@@ -26,8 +26,6 @@ class Account(Base):
     state = Column('state', String(32), nullable=True)
     roles = Column('roles', ARRAY(String), default=[], nullable=False)
 
-    UniqueConstraint(email, username, name='unique_acct_emails_usernames')
```
jtuckerk, line 29: to include this, if we decide we need it, we need to store it like this
`__table_args__ = (UniqueConstraint(email, username, name='unique_acct_emails_usernames'),)`

ecaoile: Ok. If we can do unique without UniqueConstraint, I'm fine with that.

ecaoile (approved):

PR #105, opened 2021-03-14
Replace airtable iframe with role cards

powell759:


powell759:
Noticing in the meantime we've created a standard card component.  (https://github.com/HF-RapidResponse/volunteer-portal/pull/96)

Will follow this PR with one to refactor this on that component.

client/src/components/Roles.js
```diff
@@ -1,13 +1,63 @@
...
+      >
+        <h2 className="header-3-section-lead">Oops:</h2>
+        <h2 className="header-3-section-breaker">Error loading initiatives.</h2>
+        <p>Please try again later.</p>
+      </Col>
+    );
+  }
+
+  // Build 'em if you got 'em
+  const roleCards = roles.map((role) => RoleCard(role));
```
ecaoile: RoleCard.js is yelling about needing a unique key in the console, but I think the unique key needs to be assigned here.

client/src/components/RoleCard.js
```diff
@@ -0,0 +1,25 @@
...
+import '../styles/roles.scss';
+import '../styles/bootstrap-overrides.scss';
+import {Button} from 'react-bootstrap';
+
+function RoleCard(role) {
+  return (
+    <div className="role-card">
+      <div className="card-content">
+        <h3>{role.role_name}</h3>
+        <p class="role-highlight">
```
ecaoile: needs to be className instead of class

client/src/components/Roles.js
```diff
@@ -16,17 +66,9 @@ function Roles() {
...
-        className="airtable-embed"
-        src="https://airtable.com/embed/shrZmR6ahgCsyUBGy?backgroundColor=greenLight&amp;viewControls=on"
-        frameBorder="0"
-        width="100%"
-        height="1000"
-        onLoad={() => setLoading(false)}
-        title="volunteer-listings"
-        styles="background: transparent; border: 1px solid #ccc;"
-      ></iframe>
+      <div class="role-card-container">
```
ecaoile: needs to be className instead of class

client/src/components/RoleCard.js
```diff
@@ -0,0 +1,25 @@
...
+          Priority: {role.priority}
+          <br />
+          Weekly time commitment: {role.min_time_commitment}-{role.max_time_commitment} hrs
+        </p>
+        <p class="role-overview">{role.overview}</p>
```
ecaoile: needs to be className instead of class

ecaoile (changes requested):
UI looks great. Just needs a few small changes to make React happy

client/src/components/RoleCard.js
powell759: done

client/src/components/Roles.js
powell759: done

client/src/components/RoleCard.js
powell759: done

client/src/components/Roles.js
powell759: Can you paste the error here? I'm not seeing it and not sure what you/react mean by needing a unique key.

ecaoile: works now after latest commit

ecaoile (approved):

PR #106, opened 2021-03-15
Setting up auth on cloud staging instance 

jtuckerk:
a few tweaks set up staging with earl jay's login changes -- view at https://staging-volunteer-portal-dv734ug3ua-uc.a.run.app

The one bit about this I dont like is that I needed to expose the app on port 80 for local development. somewhere in the chain of urls the port was removed so it was attempting to access http://localhost instead of localhost:8000. This is probably fine since we're serving from port 80 in prod, but I just dont like that i dont fully understand what's going on.


client/src/components/OAuth/GoogleOAuthButton.js
```diff
@@ -8,10 +8,7 @@ function GoogleOAuthButton() {
       variant="outline-primary"
       className="mt-4 mb-4"
       onClick={() => {
-        const baseUrl =
-          window.location.port === '8000'
-            ? 'http://localhost:8081'
-            : window.location.origin;
+        const baseUrl = window.location.origin;
```
ecaoile, line 11: Oh, I see it stopped being a punk on localhost! Nice!

ecaoile (approved):

PR #108, opened 2021-03-26
added auth jwt to account settings post

ecaoile:
Small kine PR. Turns out I can require auth jwt on this because the newly created user should already have a jwt token by the time the regular account post finishes running.


jtuckerk (approved):
LGTM!

PR #109, opened 2021-03-29
implement reset password functionality

ecaoile:
Figured this was simple enough to implement over a weekend.
- modal to request a password reset is on the login page (can enter e-mail or username), changes view on successful submit
- user receives e-mail with URL including a hash param to reset password (expires in 15 minutes from request)
- after entering password, client lets user know reset was success and provides a link back to the login page

### DIFF
--- a/api/account_api.py
+++ b/api/account_api.py
@@ -227,7 +227,7 @@ def create_notification(username_or_email: AcctUsernameOrEmailSchema, db: Sessio
 
             url_to_click = f'{base_url}/reset_password?hash={password_reset_hash}'
             email_message += f'<p>We have received a request to reset your password. To reset your password, \
-                please click here: <a href="{url_to_click}">{url_to_click}</a></p> \
+                please click the following link: <a href="{url_to_click}">change my password</a></p> \
                 <p>This link will expire in 15 minutes.</p>'
         else:
             email_message += f'<p>We have received a request to reset your password. \
--- a/api/account_api.py
+++ b/api/account_api.py
@@ -22,6 +22,8 @@ from security import encrypt_password, check_encrypted_password
 import re
 from datetime import datetime
 import socket
+import random
+import decimal
 
 router = APIRouter()
 
@@ -208,7 +210,7 @@ def create_notification(username_or_email: AcctUsernameOrEmailSchema, db: Sessio
             email=username_or_email.email).first()
     elif username_or_email.username is not None:
         existing_acct = db.query(Account).filter_by(
-            email=username_or_email.username).first()
+            username=username_or_email.username).first()
 
     email_message = None
     if existing_acct is not None:
@@ -218,8 +220,8 @@ def create_notification(username_or_email: AcctUsernameOrEmailSchema, db: Sessio
             acct_settings = db.query(AccountSettings).filter_by(
                 uuid=existing_acct.uuid).first()
             curr_time = datetime.now()
-            password_reset_hash = encrypt_password(
-                existing_acct.username + str(curr_time))
+            password_reset_hash = generate_str_for_hash(
+                existing_acct.username, curr_time)
             acct_settings.password_reset_hash = password_reset_hash
             acct_settings.password_reset_time = curr_time
             db.merge(acct_settings)
@@ -230,9 +232,11 @@ def create_notification(username_or_email: AcctUsernameOrEmailSchema, db: Sessio
                 please click the following link: <a href="{url_to_click}">change my password</a></p> \
                 <p>This link will expire in 15 minutes.</p>'
         else:
+            oauth_type = existing_acct.oauth.capitalize()
             email_message += f'<p>We have received a request to reset your password. \
-                Your account was created with {existing_acct.oauth.capitalize()} OAuth; therefore, \
-                you cannot set or reset a password.</p>'
+                Your account was created with {oauth_type} OAuth; therefore, \
+                you cannot set or reset a password. \
+                "Please try signing in with {oauth_type}.</p>'
         email_message += '<b> If this action was not performed by you, \
                 someone may be targeting your account. You may want to consider changing your e-mail password. </b>\
                 <p>Regards, <br/>HF Volunteer Portal Team </p>'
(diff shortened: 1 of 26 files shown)


api/account_api.py
```diff
@@ -186,3 +198,72 @@ def delete_settings(uuid, Authorize: AuthJWT = Depends(), db: Session = Depends(
...
+    email_message = None
+    if existing_acct is not None:
+        email_message = f'<p>Dear {existing_acct.first_name},</p>'
+        if existing_acct.oauth is None:
+            base_url = Config['routes']['client']
+            acct_settings = db.query(AccountSettings).filter_by(
+                uuid=existing_acct.uuid).first()
+            curr_time = datetime.now()
+            password_reset_hash = encrypt_password(
+                existing_acct.username + str(curr_time))
```
jtuckerk: We'll need to include something random in here, or just pick a random number. if someone who doesnt own the account sends a password reset request they know roughly the time and the username theyre trying to hack so just a few checks and they can get the string then use the same encrypt function since this code is public and get the reset hash

api/account_api.py
```diff
@@ -186,3 +198,72 @@ def delete_settings(uuid, Authorize: AuthJWT = Depends(), db: Session = Depends(
...
+            db.commit()
+
+            url_to_click = f'{base_url}/reset_password?hash={password_reset_hash}'
+            email_message += f'<p>We have received a request to reset your password. To reset your password, \
+                please click the following link: <a href="{url_to_click}">change my password</a></p> \
+                <p>This link will expire in 15 minutes.</p>'
+        else:
+            email_message += f'<p>We have received a request to reset your password. \
+                Your account was created with {existing_acct.oauth.capitalize()} OAuth; therefore, \
+                you cannot set or reset a password.</p>'
```
jtuckerk: Can we add, "please try signing in with  {existing_acct.oauth.capitalize()}"

client/src/components/Login.js
```diff
@@ -77,14 +81,14 @@ function Login(props) {
...
+            <div className="mt-3 mb-3">
+              <p
+                className="font-weight-light hover-hand-and-underline"
                 style={{ color: 'gray' }}
-                to="/forgot_password"
+                onClick={() => setShowModal(true)}
               >
-                Forgot Password?
-              </Link>
+                Forgot password?
```
jtuckerk: Can we make "Forgot Password?" stand out a little more as a link? underlined or centered or maybe even make it a little button? not sure whats best, but it's subtle as is
image

client/src/components/modals/ForgotPasswordModal.js
```diff
@@ -0,0 +1,82 @@
...
+            noValidate
+            validated={validated}
+            onSubmit={handleSubmit}
+            className="mt-5 mb-5"
+          >
+            <Form.Group controlId="forgot-password-request">
+              <Form.Label>Username or e-mail address</Form.Label>
+              <Form.Control
+                type="text"
+                placeholder="Enter username or e-mail address (required)"
```
jtuckerk: I dont think "(required)" is necessary since theres only one box and an error shows if its empty. it also overruns the side of the input box on some devices
image

api/account_api.py
```diff
@@ -186,3 +198,72 @@ def delete_settings(uuid, Authorize: AuthJWT = Depends(), db: Session = Depends(
...
+    existing_acct = None
+    if username_or_email.email is not None:
+        existing_acct = db.query(Account).filter_by(
+            email=username_or_email.email).first()
+    elif username_or_email.username is not None:
+        existing_acct = db.query(Account).filter_by(
+            email=username_or_email.username).first()
+
+    email_message = None
+    if existing_acct is not None:
```
jtuckerk, line 218: When walking through the password reset flow the email didn't send for my username, but did send for my email

api/account_api.py
```diff
@@ -186,3 +198,72 @@ def delete_settings(uuid, Authorize: AuthJWT = Depends(), db: Session = Depends(
...
+        email_message = f'<p>Dear {existing_acct.first_name},</p>'
```
jtuckerk, line 219: When I reset the password I got the green check marks on the new password fields, but no other indication that it was correct, nor a redirect to the login page.

I somehow got in a bad state when I continued on after resetting the password. I tried logging in using my old password first and it still worked, then I logged out and tried the new password and it didnt work. then after a bit I tried logging out and started seeing "DELETE /api/logout HTTP/1.0" 401 Unauthorized" in the api logs. so maybe the password reset took a minute and the the token got revoked? or something?

jtuckerk (changes requested):
There are a couple of funky things but overall you're continuing to kill it!

api/account_api.py
ecaoile: added random numbers

api/account_api.py
ecaoile: added sentence

client/src/components/Login.js
ecaoile: @cynthiatai Any ideas for how this should look? I was following the mock-up for this page

api/account_api.py
ecaoile: Oops, I got carried away with copypasta and searched for email twice. Fixed.

api/account_api.py
ecaoile: Hmm, I can't seem to repro this...

client/src/components/modals/ForgotPasswordModal.js
ecaoile: removed

client/src/components/Login.js
ecaoile: Cynthia said to just underline it. Is this ok?
image

jtuckerk: yes I think that helps a lot!

api/account_api.py
jtuckerk: hmm yeah me neither.

jtuckerk (approved):
It's looking good!

one last thing: I think there's a validity check mismatch for usernames, I can't tell quite where since the validity checking syntax is confusing to me, but in creating an account it lets me create tuck, but when I go to use that to retrieve my reset it triggers the not >4 char error

jtuckerk:
ooh also I think this is more of a feature than a bug, but im not sure if it's intential. 
I can create an account using the basic account creation, and then later log in with oauth for that email. it saves things like my original username so that's good. would probably preserve my initiative preferences too. My only worry would be something like the "oauth" field in the account db entry being wrong and later causing issues. maybe for password reset or something. 

also not added in this PR.  just noting it here.

ecaoile:
> 
> 
> ooh also I think this is more of a feature than a bug, but im not sure if it's intential.
> I can create an account using the basic account creation, and then later log in with oauth for that email. it saves things like my original username so that's good. would probably preserve my initiative preferences too. My only worry would be something like the "oauth" field in the account db entry being wrong and later causing issues. maybe for password reset or something.
> 
> also not added in this PR. just noting it here.

I noticed that too. I think if you oauth login an existing account that was created with basic auth, it leaves the oauth field as null. That means you'll still be allowed to reset password. I don't know if we wanted to disable it...but I think we could maybe add some stuff like adding the oauth profile pic if they don't already have one

PR #111, opened 2021-04-03
move events list above calendar on events calendar page

ecaoile:
Addressing this task: https://app.asana.com/0/1196959085120745/1200090551718174/f

Also accounted for edge cases where you don't have any upcoming events:
image


client/src/components/Calendar.js
```diff
@@ -26,46 +26,52 @@ function Events() {
...
+              </Col>
+            );
+          } else {
+            return null;
+          }
         }
-      }
-    );
+      )
+      .filter((event) => !!event);
```
jchi2241, line 66: did we verify that this fixes the "no events" edge case that was breaking the app? iirc react just ignores falsey values in an array. e.g., rendering
```
[null, <Col>...</Col>, null]
```
is the same as 
```
[<Col>...</Col>]
```

could the issue that was breaking the app be that the `data` returned was not an array, but `undefined` or some object that doesn't have the `.map` function?

ecaoile: Yes, I console logged it myself. There were was an array of 14 nulls being returned.

jchi2241: sorry, what i meant was react shouldn't be erroring when rendering `null`, so filtering out those null values probably didn't fix the crashing

the issue is more likely that `data` either 
- can't be mapped on. e.g., `data` is undefined
- each item in `data` isn't an object, so attempting to destructure those keys (uuid, event_name, etc...) is erroring
- each item in `data` IS an object, but the destructured key doesn't exist, so its undefined, and passing undefined into `moment("undefinedZ")` and calling startDate.isAfter() is erroring.

ecaoile: Oh, I was only attempting to address an array of nulls showing up because the header for the upcoming events was showing up but there were no message saying there weren't any upcoming events. I do notice that the error message pops up when I disconnect locally, but I figured that was intended and that useQuery was working like an event listener.

I also see this in the console logs:
image

I'm not sure if this is what you're talking about, but it appears in the volunteer openings page too. There's some weird behavior going on with airtable.

jchi2241: > Oh, I was only attempting to address an array of nulls showing up because the header for the upcoming events was showing up but there were no message saying there weren't any upcoming events.

ah i see... yea what you have should fix that then. 

> I'm not sure if this is what you're talking about, but it appears in the volunteer openings page too. There's some weird behavior going on with airtable.

i was referring to the crashing that @jtuckerk pointed out awhile back that happened on the events page in staging, and there was some convo about the reason being the db in staging wasn't being populated with the expected event info or something.

no idea what's causing that console error :/

jchi2241 (approved):

PR #112, opened 2021-04-05
Move some account logic to the backend. Add Account api tests

jtuckerk:
Major changes:
- make account settings creation and deletion both happen within the account creation/deletion api call. 
account creation now happens in a single client call -- easier to test (at least for me) and seems less likely to end up in a bad state with a partially created account. should be faster too. 
- changed the account schema a little bit to only require fields necessary for a call. This was going to be part of a bigger change to make it so the client could update a single field in the settings, for example, without needing to send all of the fields to the api. this was going to take more work than I first realized and isnt worth it. I still like the schema simplification tho.
- test


ecaoile (approved):
Looks like everything is working properly

PR #113, opened 2021-04-06
Order events in events api call and in initiatives.

jtuckerk:


ecaoile (approved):
dates check out on the API endpoint

PR #114, opened 2021-04-06
replaced under construction display for data page

ecaoile:
image

We'd have to remove the container wrapper along with the grid if we wanted to dynamically make the page content wider on larger resolutions like the [mock-up for this page.](https://www.figma.com/file/W84pQJImct4asHoqmWGKoT/HFRR-Volunteer-Portal?node-id=1540%3A25) Do we want to do that?


jtuckerk (approved):
The resizing all looks fine to me, I dont think we need to change it

PR #116, opened 2021-04-07
fixed oauth users not getting settings created on registration

ecaoile:
Follow-up to #112 

Settings were turning up null. The involvement page wasn't properly handling empty settings either.


jtuckerk (approved):
ahh good catch, sorry about that.

PR #117, opened 2021-04-19
Adding account registration confirmation e-mail for basic auth accounts

ecaoile:
Requiring users to open an e-mail and confirm registration if they didn't log in with OAuth. This shouldn't affect OAuth users at all.

### DIFF
--- a/api/account_api.py
+++ b/api/account_api.py
@@ -51,7 +51,7 @@ def create_access_and_refresh_tokens(user_id: str, Authorize: AuthJWT):
 
 def check_valid_password(password: str):
     match = re.search(
-        "^.*(?=.*\d)(?=.*[a-zA-Z])(?=.*\d)(?=.*[`!@#$%^&*()_+\-=[\]{};':\"\\|,.<>/?~]).{6,20}$", password)
+        r"^.*(?=.*\d)(?=.*[a-zA-Z])(?=.*\d)(?=.*[`!@#$%^&*()_+\-=[\]{};':\"\\|,.<>/?~]).{6,20}$", password)
     if match is None:
         raise HTTPException(status_code=400,
                             detail=f"Please enter a password between 6 and 20 characters long with at least 1 letter, 1 number, and 1 special character.")
--- a/api/account_api.py
+++ b/api/account_api.py
@@ -261,9 +261,11 @@ def create_email(notification_payload: AccountNotificationSchema, existing_acct:
             acct_settings.verify_account_hash = verify_account_hash
             db.merge(acct_settings)
             db.commit()
-            email_message = f'<p>Dear {existing_acct.first_name},</p>'
+
             verify_url = f'{base_url}/verify_account?hash={verify_account_hash}'
             cancel_url = f'{base_url}/cancel_registration?hash={verify_account_hash}'
+
+            email_message = f'<p>Dear {existing_acct.first_name},</p>'
             email_message += f'<p>We have received a request to create an account. Accounts not created with OAuth require \
                         e-mail verification. Please click the following link: <a href="{verify_url}">verify my account</a></p>'
             email_message += f'<b>If this action was not performed by you or performed by accident, \
@@ -293,6 +295,25 @@ def get_settings_from_hash(pw_reset_hash: str, Authorize: AuthJWT = Depends(), d
     else:
         raise HTTPException(status_code=400,
                             detail=f"Invalid or expired password reset URL!")
+
+
+@router.get("/verify_account_from_hash", status_code=200)
+def complete_account_registration(verify_hash: str, Authorize: AuthJWT = Depends(), db: Session = Depends(get_db)):
+    existing_settings = db.query(AccountSettings).filter_by(
+        verify_account_hash=verify_hash).first()
+    if existing_settings is not None:
+        existing_settings.verify_account_hash = None
+        db.merge(existing_settings)
+        db.commit()
+        existing_account = db.query(Account).filter_by(
+            uuid=existing_settings.uuid).first()
+        if existing_account is not None:
+            existing_account.is_verified = True
+            db.merge(existing_account)
+            db.commit()
+    else:
+        raise HTTPException(status_code=400,
+                            detail=f"invalid hash or account does not exist")
 
 
 def generate_str_for_hash(username: str, curr_time: datetime):
--- a/api/account_api.py
+++ b/api/account_api.py
@@ -216,37 +216,61 @@ def create_notification(notification_payload: AccountNotificationSchema, db: Ses
     elif notification_payload.username is not None:
         existing_acct = db.query(Account).filter_by(
             username=notification_payload.username.strip()).first()
+    create_email(notification_payload, existing_acct, db)
 
+
+def create_email(notification_payload: AccountNotificationSchema, existing_acct: Account, db: Session):
     email_message = None
     if existing_acct is not None:
-        email_message = f'<p>Dear {existing_acct.first_name},</p>'
-        if existing_acct.oauth is None:
+        if notification_payload.notification_type == 'password_reset':
+            email_message = f'<p>Dear {existing_acct.first_name},</p>'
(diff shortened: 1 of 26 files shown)


jtuckerk (commented):
I'm not sure if this is ready to go or not yet, but I tested it!
I like the option to undo the account creation, thats a good feature!

Is it possible/ easy to skip the login step after the verification? like have the verify account link send them a token and land them at the account settings page? having to retype my email and password is a buzzkill

ecaoile:
> 
> 
> I'm not sure if this is ready to go or not yet, but I tested it!
> I like the option to undo the account creation, thats a good feature!
> 
> Is it possible/ easy to skip the login step after the verification? like have the verify account link send them a token and land them at the account settings page? having to retype my email and password is a buzzkill

I was going to write tests before having someone review it and also wanted to verify the current tests were working on here. Almost done...but yes, I can generate a token and have them log back in. Will probably set a 3 second timer so they have time to read the text.

api/tests/test_account_api.py
```diff
@@ -266,36 +277,83 @@ def test_reset_account_password_with_hash(db):
...
+
+    # notification for
+    resp = client.post(f'api/notifications/',
+                       json={"username": "DakotaMcclain", "email": "rebecca03@thomasrivera.com", "notification_type": "verify_registration"})
+    assert resp.status_code == 204
+
+    resp = client.get(f'api/account_settings/{uuid}')
+    assert resp.status_code == 200
+
+    # settings = resp.json()
```
ecaoile: @jtuckerk @trippersham need help getting this commented out part to work. It can't seem to read the verify_account_hash and the cancel_registration_hash that I added recently. I don't know how to update the test API to include the new props, and I don't know how to get the tests working locally either.

jtuckerk: you were just missing the fields in the API schema. It was only a matter of time before that bit someone in the but. I didn't like having essentially duplicate definitions for the db model and the API schema, but we do need that separation so we can have requests/responses with a subset of the db fields. I think for now it's a necessary evil.

 I've pushed a change that adds the fields and fixes the tests

jtuckerk: On second thought we probably don't need to expose those through the API. I dont think there is any harm in it so I might just leave it.

If you think of some way that could be insecure, we can update the tests to access the DB directly to check that those fields are populated.

ecaoile: Oh silly me, I updated it for password reset hash but totally forgot it for verify. RIP

I don't think exposing hashes through the API is too big of a deal. I mean there's inherently going to be a potential security issue with allowing people to verify through a URL...but we want them to be able to register through a URL.

api/account_api.py
```diff
@@ -204,84 +193,3 @@ def update_settings(uuid, settings: AccountSettingsSchema, Authorize: AuthJWT =
...
-        str(curr_time) + str(random.uniform(1, one_rand_num)
-                             + random.uniform(1, one_rand_num) + random.uniform(1, one_rand_num)))
-    return password_reset_hash
-
-
-def minutes_difference(reset_req_time: datetime):
-    curr_time = datetime.now()
-    seconds_diff = abs((curr_time - reset_req_time).seconds)
-    minutes_diff = seconds_diff/60
-    return minutes_diff
```
jtuckerk, line 287: Breaking all of this into its own file was the right call :)

api/auth.py
```diff
@@ -120,12 +121,16 @@ def authorize_basic(account: AccountBasicLoginSchema, Authorize: AuthJWT = Depen
     existing_acct = db.query(Account).filter_by(email=account.email).first()
     if existing_acct is None:
         raise HTTPException(
-            status_code=400, detail=f"An account with the email address {account.email} does not exist")
+            status_code=403, detail=f"E-mail or password is invalid!")
```
jtuckerk, line 124: +1

api/notifications_api.py
```diff
@@ -0,0 +1,174 @@
...
+@router.delete("/cancel_registration", status_code=204)
+def delete_account_from_hash(cancel_hash: str, Authorize: AuthJWT = Depends(), db: Session = Depends(get_db)):
+    settings_to_delete = db.query(AccountSettings).filter_by(
+        cancel_registration_hash=cancel_hash).first()
+    if settings_to_delete is None:
+        raise HTTPException(status_code=400,
+                            detail=f"Account settings with hash {cancel_hash} not found")
+    acct_uuid = settings_to_delete.uuid
+    acct_to_delete = db.query(Account).filter_by(uuid=acct_uuid).first()
+    db.delete(settings_to_delete)
```
jtuckerk: just to be safe, can we call the same method as the API calls to delete the account? we may need to pull the account delete function out into one more level of function calls since I'm not sure how/if we should call the router decorated function directly.
I think for now this is logically correct since it cleans up the account and the settings dbs, but in the future if we do things like remove a volunteer or admin from a group DB we can do that all from the same function and not have to update it here and elsewhere.

api/tests/test_account_api.py
```diff
@@ -266,36 +277,87 @@ def test_reset_account_password_with_hash(db):
...
+        f'api/accounts/{uuid}', json={'password': 'new_password.123'})
     assert resp.status_code < 400
 
     new_client = TestClient(app)
-    resp = new_client.post(f'api/auth/basic', json={'email':'rebecca03@thomasrivera.com',
-                                                        'password': 'wrong_password.123'})
+    resp = new_client.post(f'api/auth/basic', json={'email': 'rebecca03@thomasrivera.com',
+                                                    'password': 'wrong_password.123'})
     assert resp.status_code > 400
-    resp = new_client.post(f'api/auth/basic', json={'email':'rebecca03@thomasrivera.com',
```
jtuckerk, line 296: what tools are you using to auto format? I'll try to install the same one/similar and maybe include it in the README so we can all be formatting the same way and avoid some noise in these reviews

client/src/store/user-slice/index.js
```diff
@@ -231,17 +232,19 @@ export const attemptRegister = (payload) => async (dispatch) => {
...
-    dispatch(setUser(user));
-    return;
+    const createdAcct = accountRes.data;
+    if (createdAcct) {
+      const obj = {
+        notification_type: 'verify_registration',
+        email: createdAcct.email,
+        username: createdAcct.username,
+      };
+      await axios.post(`/api/notifications/`, obj);
```
jtuckerk, line 242: No need to update this now, but my leaning is to run things like this from within the create call in the api just to avoid a possible failure between http calls

jtuckerk (changes requested):
only 1 thing I think it's worth changing

api/tests/test_account_api.py
ecaoile: I have ESLint and Prettier installed in VS Code. That might be redundant, but it is reading off our eslintrc.json file in the client. ESLint is also mentioned in the extensions section of setup.md.

api/notifications_api.py
ecaoile: I thought about that, but the reason I didn't do it was because the client doesn't have a token at that point. I guess we can inject a token creation, but that seems a little weird since the user is going to disappear right away.

client/src/store/user-slice/index.js
ecaoile: Hmmm, I have mixed feelings on that. The notification is outside the scope of accounts. Something feels dirty about doing that, but yes, we can reduce the number of calls by combining stuff

api/notifications_api.py
ecaoile: Ok, I created a helper function that contains both deletes and separated from the API call so that any other function can call it. We can just have the access check happen at the beginning of the API call function before calling the helper that does the actual deleting.

api/account_api.py
ecaoile: kthx

api/auth.py
ecaoile: +2

jtuckerk (approved):

PR #120, opened 2021-05-01
Create emails templating helper functions and spice up emails.

jtuckerk:
Trying to reference our style sheets from the api to keep the client and the email look and feel in sync was going to be too hard, so I've just manually created some templates within the api based of the client style.

I made some slight changes to the wording of some of the messages.

Styling is more limited on some mail clients than others, so in gmail for example you wont see the drop shadow

### DIFF
--- a/api/notifications_api.py
+++ b/api/notifications_api.py
@@ -227,7 +227,7 @@ BUTTON_TEMPLATE = f"""<tr><td align="center">
 text-align: center; vertical-align: middle; -webkit-user-select: none; user-select: none; border:
 1px solid transparent; padding: .375rem .75rem; font-size: 1rem; line-height: 1.5;
 border-radius: .25rem; color: #fff; background-color: {button_color}; border-color: {button_color};
-text-decoration: none;">
+text-decoration: none; margin: 1rem auto;">
 {{button_text}}
 </a></td></tr>"""
 
--- a/api/notifications_api.py
+++ b/api/notifications_api.py
@@ -55,19 +55,23 @@ def create_and_send_email(notification_payload: AccountNotificationSchema, exist
                 db.merge(acct_settings)
                 db.commit()
 
-                url_to_click = f'{base_url}/reset_password?hash={password_reset_hash}'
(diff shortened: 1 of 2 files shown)


jtuckerk:
image
image
image

jtuckerk:
@trippersham for reference

api/notifications_api.py
```diff
@@ -170,3 +181,58 @@ def minutes_difference(reset_req_time: datetime):
...
+                </table>
+            </td>
+        </tr>
+    </table>
+</body>
+"""
+
+# Try to keep this in sync with the style from the CSS element .btn and .btn-info in the client
+BUTTON_TEMPLATE = f"""<tr><td align="center">
+<a href="{{link}}" style="display: inline-block; font-weight: 400; color: #212529;
```
ecaoile, line 226: I'd personally add at least 1rem vertical margins on the button. Things looked slightly squished. Think something like `margin: 1rem auto` as one of the style props.

api/notifications_api.py
```diff
@@ -170,3 +181,58 @@ def minutes_difference(reset_req_time: datetime):
     seconds_diff = abs((curr_time - reset_req_time).seconds)
     minutes_diff = seconds_diff/60
     return minutes_diff
+
+background_color = "#f3f2ef"
+card_background_color = "#ffffff"
+button_color = "#399DAC"
+# For email header image
+base_url = Config['routes']['client']
+if 'localhost' in base_url:
```
ecaoile, line 190: I'd personally do a ternary equivalent, but if you want the localhost reasoning to stick out, this is fine too.

ecaoile (approved):
Styling looks fabulous although I'd like some vertical spacing with the buttons.

api/notifications_api.py
jtuckerk: I think I do want it to stick out more. I like how it reads this way. I should use the ternary operator more, but today isnt the day im gonna start

api/notifications_api.py
jtuckerk: good call. done

PR #123, opened 2021-05-12
decoupled identifiers and verification

jtuckerk:
Tripps Changes:
  * Adds PersonalIdentifiers to capture sensitive, unique IDs
  * Adds VerificationTokens to handle identifier verification lifecycle
  * Adds APIs for beginning and finishing identifier verification
  * Email verification emails triggered automatically upon registration
  *  Support for account and non-account registrations
  * Tests and various improvements

Integrating with existing things:
  * Tried to keep as many of the api routes and schemas the same
  * Added some db relationships such as "email" to pull the email from an identifier to mimic the old email column behavior
  * removed hash verification in favor of Tripps code that was already set up with the personal identifiers
  * modified the personal identifiers verification to work for the account-less subscription flow (coming soon)
  * Added some "cascade" specifiers to clean up account_settings, identifiers, and verification tokens when accounts are deleted
  * Added some Hash-based indexes to some tables to make lookups fast and efficient

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -105,11 +105,10 @@ def oauth_login_or_creation(id_type,
                             identifier_class=PersonalIdentifier,
                             city=None,
                             state=None):
-    print("AAAAAAAAAAAAAAAA", identifier_value)
     id_state, identifier = get_or_create_identifier(id_type,
                                                     identifier_value, db,
                                                     identifier_class)
-    print(identifier)
+
     if id_state is IdentifierState.NONE_MATCHING \
        or id_state is IdentifierState.NO_ACCOUNT_IS_VERIFIED \
        or id_state is IdentifierState.NO_ACCOUNT_NOT_VERIFIED:
--- a/api/auth.py
+++ b/api/auth.py
@@ -8,17 +8,19 @@ from authlib.integrations.starlette_client import OAuth, OAuthError
 from fastapi import HTTPException
 from fastapi.responses import HTMLResponse, RedirectResponse
 from fastapi_jwt_auth import AuthJWT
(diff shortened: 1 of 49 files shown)


api/auth.py
```diff
@@ -77,24 +84,56 @@ async def authorize_google(request: Request, Authorize: AuthJWT = Depends(), db:
...
-            last_name=user.family_name,
-            oauth='google',
-            profile_pic=user.picture,
-            is_verified=True,
-        )
+            username=username,
+            first_name=first_name,
+            last_name=last_name,
+            profile_pic=profile_pic,
+            primary_email_identifier=primary_email,
```
ecaoile, line 124: You've removed the oauth parameter that was previously present. This is allowing oauth users to change their password, which doesn't really make sense if they don't have a password to begin with.

jtuckerk: ah good catch. I'll fix that to check the type of identifiers they have connected to the account

jtuckerk: oh wait, I did that already. see line 59 of api/notifications_api.py

ecaoile: You created the new prop on the back end, but the front end isn't reading it. Try reading the new prop on the front end (settings.js) instead of the oauth prop.

jtuckerk: ah gotchya, I missed the option in the settings to change the password. I was only looking at the "forgot password" reset flow.
I've introduced a primary_identifier_type field into the account to show the IdentifierType from personal_identifiers.py.
Right now I'm just assuming that an email identifier (including google oauth) is the primary identifier, otherwise the first identifier found for the account is the primary.

PR #124, opened 2021-05-15
Gracefully handle missing image on events

ecaoile:
Addressing the Asana task with the [same name](https://app.asana.com/0/1196959085120745/1199354681082661/f)

I'm feeling a little brain-dead on how to write tests here...


jtuckerk (approved):
Looks good! I added a test

I dont know what that image is, but it doesnt seem to be public. We can add an HF image in later

jtuckerk:
nvm the image is an HF one, just cant open it directly in a browser or something

PR #125, opened 2021-05-26
Add subscriptions functionality for initiatives in the API

jtuckerk:
This sets up the subscriptions db so they can be applied for more than just initiatives, but only the API has been implemented for initiatives.
Subscriptions have an associated identifier and account if there is an account associated to the identifier. If a user is logged in their subscriptions are automatically verified. if no user is logged in, subscribing on the initiatives pages triggers an email notification to verify the subscription.

To simplify things, I started to update the client to change from email submission forms to toggle switches on the initiative cards, but did not finish this (and what I did do is bad).
I have not made any changes to the accounts settings for the initiative map, but we should remove the initiative_map from the settings and use the `api/subscriptions/account/{uuid}/initiative_map` endpoint or `api/subscriptions/account/{uuid}/initiatives`

The last thing is that the redirect stuff is not set up. after verifying a token it uses the same flow for subscriptions and identifiers.

image
This image just shows setting up the subscription feature on the cards based on the index, and not the user


jtuckerk:
@ecaoile think you could help me finish up the client side work for this? I really struggle as JS

ecaoile:
> 
> 
> @ecaoile think you could help me finish up the client side work for this? I really struggle as JS

What were you looking to implement on the client besides the toggle switches? I see there's one switch, but it doesn't toggle on/off. I'm guessing we want to remove the initiative map and use that new endpoint.

jtuckerk:
> > @ecaoile think you could help me finish up the client side work for this? I really struggle as JS
> 
> What were you looking to implement on the client besides the toggle switches? I see there's one switch, but it doesn't toggle on/off. I'm guessing we want to remove the initiative map and use that new endpoint.

yeah finishing up the switches on the initiative page, the switches on the settings page, and some sort of landing page/indicator when the subscription is verified

api/subscriptions_api.py
```diff
@@ -0,0 +1,153 @@
...
+            title=f'Please verify your subscription with {Config["public_facing_org_name"]}')
+
+    else:
+        raise HTTPException(status_code=400, detail="No identifier provided")
+
+    db.add(sub)
+    db.commit()
+    return {'uuid': sub.uuid}
+
+
```
ecaoile, line 98: Some of these endpoints hurt my feelings. Why is the post for adding a subscription using the initiative UUID, but the unsubscribe is directly using the subscription UUID? Are we supposed to have access to the subscription UUID on the front end? Seems to me you'd have to use perform a get all just to look for the matching subscription when deleting/unsubscribing from an initiative

ecaoile:
@jtuckerk can you work with this? I got the initiatives toggles on the main page synced up with user and displaying the same values as the ones on the involvement page. I also have the main initiatives page checking for a logged in user so that it can switch between displaying the toggle switch and manually entering their email.

Also, the delete subscription endpoint was being a punk, so I had it filter the sub by initiative uuid. Even when I got what I thought was the sub uuid, it never ended up matching any existing uuids. I dunno what happened there. Test might need to be refactored.

ecaoile (changes requested):
Please review my changes.

api/subscriptions_api.py
jtuckerk: We should always have the subscription UUID on the frontend since we'll need to get the subscription state for a user in order to set up the toggles, then we can just hang on to the subscription UUID.

jtuckerk:
> @jtuckerk can you work with this? I got the initiatives toggles on the main page synced up with user and displaying the same values as the ones on the involvement page. I also have the main initiatives page checking for a logged in user so that it can switch between displaying the toggle switch and manually entering their email.
> 
> Also, the delete subscription endpoint was being a punk, so I had it filter the sub by initiative uuid. Even when I got what I thought was the sub uuid, it never ended up matching any existing uuids. I dunno what happened there. Test might need to be refactored.

I can definitely work with this! At the time I thought it would be a good idea to keep the same structure with the initiative map, but in hindsight I think that's what's causing the problem. It's weird since if a toggle is on there is a corresponding subscription and valid UUID, but if it's off that's just an implicit signal that no subscription is present.

jtuckerk:
@ecaoile I think it's done. The subscription API was definitely shitty. It's a little better now. It's working and I dont think it needs to be perfect.

